PR #30, opened 2026-01-14
long tail admin apis

airhorns:
- **Return heartbeat result**
- **Add support for worker side cancellation**
- **Add API for retrieving attempts with job at the same time**


airhorns:

- #30 ◀
    - #31

PR #88, opened 2026-02-04
Implement lazy per-queue concurrency hydration

airhorns:
Remove synchronous full scan of concurrency holders at shard startup. Instead, hydrate queues lazily on first try_reserve() call, scanning only the specific tenant+queue combination needed. This improves startup performance when there are many active jobs.

Key changes:
- Add hydrated_queues tracking to ConcurrencyCounts
- Add hydrate_queue() for per-queue prefix scanning
- Make try_reserve() async with db/range params for lazy hydration
- Add try_reserve_sync() helper for unit tests
- Remove startup hydration call from JobStoreShard::new()
- Add integration tests for lazy hydration correctness

### DIFF
--- a/src/concurrency.rs
+++ b/src/concurrency.rs
@@ -33,19 +33,62 @@ use std::sync::Mutex;
 
 use slatedb::{Db, DbIterator, WriteBatch};
 
-use crate::dst_events::{self, DstEvent};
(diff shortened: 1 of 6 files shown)


airhorns:
Re-running CI to check for flakiness

PR #124, opened 2026-02-13
Implement job reimport for migration support

airhorns:
## Summary

- Add reimport capability: when importing a job that already exists, validate and apply new attempts, clean up old scheduling state, and update status accordingly
- Enables incremental migration from external job systems where jobs are repeatedly reimported as the external system processes them
- Strict validation ensures attempt consistency (status type, timestamps, error_code/result bytes must match existing data)
- Job info (payload, metadata, limits, priority) is NOT updated on reimport — only attempts and status change
- All Alloy sigils synchronized: `[SILO-IMP-*]`, `[SILO-REIMP-*]`, `[SILO-IMP-CONC-*]`, `[SILO-REIMP-CONC-*]`

## Test plan

- [x] 30 new reimport tests: state transitions (12), precondition rejection (4), scheduling cleanup (4), counters (4), multi-reimport (2), interaction (2), data integrity (2)
- [x] All 61 import tests pass
- [x] Full test suite passes (`cargo test -- --skip k8s_ --skip etcd_ --skip coordination_split_tests`)
- [x] Sigil validation passes (`node scripts/validate-silo-sigils.js`)
- [ ] CI green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/job_store_shard/import.rs
+++ b/src/job_store_shard/import.rs
@@ -3,9 +3,9 @@
 //! Allows importing jobs from other systems with historical attempt records.
 //! Unlike enqueue, import accepts completed attempts and lets Silo take ownership going forward.
 
+use slatedb::IsolationLevel;
 use slatedb::config::WriteOptions;
-use slatedb::{IsolationLevel, WriteBatch};
-use tracing::{debug, warn};
+use tracing::debug;
 use uuid::Uuid;
 
 use crate::codec::{decode_attempt, decode_task, encode_attempt, encode_job_info};
@@ -576,8 +576,16 @@ impl JobStoreShard {
             } else {
                 // No task found - TicketRequested case (request record exists but no task
                 // in DB queue). Delete concurrency requests for this job.
-                self.delete_concurrency_requests_for_job(&txn, tenant, job_id, &stored_limits)
-                    .await?;
+                self.delete_concurrency_requests_for_job(
+                    &txn,
+                    tenant,
+                    job_id,
+                    &stored_limits,
+                    attempt_number,
+                    start_time_ms,
+                    priority,
+                )
+                .await?;
             }
         }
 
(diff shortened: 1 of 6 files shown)


src/job_store_shard/import.rs
```diff
@@ -267,6 +309,398 @@ impl JobStoreShard {
...
+        let (new_job_status, status_kind, is_terminal) =
+            determine_import_status(params, total_attempts, now_ms, effective_start_at_ms);
+
+        // === Clean up old scheduling state ===
+
+        // [SILO-REIMP-5] Remove DB queue tasks for this job
+        let task_group = existing_job.task_group().to_string();
+        let task_prefix = task_group_prefix(&task_group);
+        let task_end = end_bound(&task_prefix);
+        let mut task_iter = txn.scan::<Vec<u8>, _>(task_prefix..task_end).await?;
```
airhorns: this is a potentially very expensive scan for every job imported which is prohibitively expensive. i think that it should be possible to find and compute the task keys for jobs in every state -- we need to do that on cancellation already, no? see what we already do and if we can re-use that. before making changes, shore up the assertions in the tests to make sure we robustly assert tasks are removed (like say by counting outstanding tests before and after each import)

src/job_store_shard/import.rs
```diff
@@ -267,6 +309,398 @@ impl JobStoreShard {
...
+        let mut grants = Vec::new();
+        if !is_terminal {
+            let next_attempt = total_attempts + 1;
+            let new_task_id = Uuid::new_v4().to_string();
+
+            // [SILO-REIMP-8] status = Scheduled
+            // [SILO-REIMP-9] new task in DB queue, old tasks for j removed
+            // [SILO-REIMP-CONC-1/2] concurrency granted path
+            // [SILO-REIMP-CONC-3/4] concurrency queued path
+            let mut writer = TxnWriter(&txn);
```
airhorns: is it incorrect to create more than one TxnWriter for the same transaction?

PR #136, opened 2026-02-18
Add toxiproxy gRPC client integration tests

kirinrastogi:
## Summary
- Add toxiproxy to the dev environment (nix devshell, process-compose) for network fault injection testing
- Instead of configuring all inter cluster communication to go through toxiproxy via setting advertised ips to be toxiproxy, I added addressMap to the client for test purposes. - if we want to also test inter cluster communication via toxiproxy we can change this
- Add 13 integration tests covering latency, bandwidth limits, timeouts, TCP slicing, proxy disable/enable, partial failures, combined toxics, and topology refresh under network issues

## Test plan
- [x] All 13 new toxiproxy integration tests pass locally
- [x] All 49 existing client integration tests pass (both direct and proxied)
- [x] All 28 worker integration tests pass
- [x] Linting passes
- [x] CI passes with toxiproxy running via process-compose

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -221,7 +221,7 @@ jobs:
         shell: bash
         run: |
           process-compose up --tui=false &
-          sleep 2
+          sleep 5
       - name: Build typescript package
         shell: bash
         working-directory: typescript_client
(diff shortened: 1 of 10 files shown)


kirinrastogi:
going to merge this, the only controversial thing may be the addressMap and replacement in the client, but I think it is necessary for the client to use a proxy - it is better than sending all cluster communication through toxiproxy.

PR #138, opened 2026-02-19
Add shared-secret Bearer token auth for gRPC

kirinrastogi:
## Summary

- Add optional `auth_token` field to `ServerConfig` — when set, a server-side interceptor rejects any gRPC request missing a valid `authorization: Bearer <token>` metadata header. Health check and reflection services remain unauthenticated so k8s probes and debugging tools still work.
- Add `AuthInterceptor` on the client side (`cluster_client.rs`, `cluster_query.rs`) so that node-to-node and WebUI remote calls automatically include the token when configured.
- Add `--auth-token` global flag to `siloctl` (also settable via `SILO_AUTH_TOKEN` env var) so operators can run commands against authenticated production clusters.
- Add `token` support for non-TLS connections in the TypeScript client (previously only worked with TLS).
- Add gRPC auth secret and wiring for the local OrbStack k8s test deployment.
- Auth is **off by default** (`auth_token` defaults to `None`); existing configs and tests are unaffected.
- Uses dev/debug cargo profile for ci and orbstack now

## Changes

| File | Change |
|------|--------|
| `src/settings.rs` | Add `auth_token: Option<String>` to `ServerConfig` |
| `src/server.rs` | Server-side interceptor validates `authorization: Bearer <token>` |
| `src/cluster_client.rs` | `AuthInterceptor` struct, `InterceptedSiloClient` type alias, `auth_token` on `ClientConfig` |
| `src/cluster_query.rs` | Thread auth token through remote SQL query path |
| `src/webui.rs` | Pass auth token to `ClusterClient` and `ClusterQueryEngine` |
| `src/siloctl.rs` | Add `auth_token` to `GlobalOptions`, wire into `connect()` |
| `src/bin/siloctl.rs` | Add `--auth-token` CLI flag with `SILO_AUTH_TOKEN` env var support |
| `Cargo.toml` | Enable clap `env` feature for env var support |
| `typescript_client/src/client.ts` | Fix `token` option for non-TLS connections via `RpcInterceptor` |
| `example_configs/k8s-gcs.toml` | Add commented `auth_token` example |
| `deploy/local-test/secret.yaml` | New k8s Secret with gRPC auth token |
| `deploy/local-test/statefulset.yaml` | Wire `SILO_AUTH_TOKEN` env var from secret into config |
| `tests/grpc_auth_tests.rs` | 5 new integration tests |

## Usage

**Server config** (`k8s-gcs.toml`):
```toml
[server]
auth_token = "${SILO_AUTH_TOKEN}"
```

**siloctl**:
```sh
siloctl --auth-token <token> cluster info
# or
SILO_AUTH_TOKEN=<token> siloctl cluster info
```

**TypeScript client**:
```ts
const client = new SiloGRPCClient({
  servers: "silo:50051",
  token: "my-secret-token",
  useTls: false,  // token now works with non-TLS too
});
```

## Manual testing (tophat)

Deployed to local OrbStack k8s cluster with gRPC auth enabled, then verified:

1. **grpcurl**: Port-forwarded to the silo service and called `GetClusterInfo` with no auth — received `UNAUTHENTICATED: missing authorization header`. Retried with `-H "authorization: Bearer grpc-auth-token"` — succeeded and returned full cluster info.

2. **TypeScript worker**: Ran a script that creates a `SiloGRPCClient` without a token — `refreshTopology()` fails with "Failed to refresh cluster topology from any server". Then created a client with `token: "grpc-auth-token"`, enqueued a job, started a `SiloWorker`, and the job was successfully picked up and processed.

3. **siloctl**: Ran `siloctl --address http://localhost:50051 cluster info` without auth — received `Unauthenticated: missing authorization header`. Ran with `--auth-token grpc-auth-token` — succeeded. Also verified `SILO_AUTH_TOKEN=grpc-auth-token siloctl cluster info` works via env var.

## Test plan

- [x] 5 new integration tests in `tests/grpc_auth_tests.rs`:
  - Unauthenticated client rejected when auth is enabled
  - Wrong token rejected
  - Correct token accepted
  - Unauthenticated client works when auth is disabled (no regression)
  - Health check bypasses auth
- [x] All existing Rust test suites pass (cluster_client, siloctl, webui, grpc_*, turmoil)
- [x] All 197 TypeScript client integration tests pass unchanged
- [x] Manual tophat against local OrbStack k8s deployment (see above)

### DIFF
--- a/nix/modules/rust.nix
+++ b/nix/modules/rust.nix
@@ -49,15 +49,29 @@
       
       # Build dependencies separately for caching
       cargoArtifacts = craneLib.buildDepsOnly commonArgs;
-      
+
       # Main package
       silo = craneLib.buildPackage (commonArgs // {
         inherit cargoArtifacts;
         doCheck = false; # Tests require external services, CI runs them separately
       });
+
+      # Debug build (fast to compile, no optimizations)
+      debugArgs = {
+        inherit src;
+        strictDeps = true;
+        nativeBuildInputs = [ pkgs.protobuf pkgs.flatbuffers ];
+        CARGO_PROFILE = "dev";
+      };
+      cargoArtifactsDebug = craneLib.buildDepsOnly debugArgs;
+      silo-debug = craneLib.buildPackage (debugArgs // {
+        cargoArtifacts = cargoArtifactsDebug;
+        doCheck = false;
+      });
     in
     {
       packages.default = silo;
       packages.silo = silo;
+      packages.silo-debug = silo-debug;
     };
 }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ version = "0.1.0"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-clap = { version = "4.3.14", features = ["derive"] }
+clap = { version = "4.3.14", features = ["derive", "env"] }
 futures = "0.3"
 slatedb = { git = "https://github.com/slatedb/slatedb.git", branch = "main" }
 object_store = { version = "0.12", features = ["gcp"] }
--- a/deploy/local-test/secret.yaml
+++ b/deploy/local-test/secret.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: silo-auth
+  namespace: silo-test
+type: Opaque
+stringData:
+  token: "grpc-auth-token"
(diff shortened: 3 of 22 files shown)


nix/modules/rust.nix
```diff
@@ -49,15 +49,29 @@
...
-      
+
       # Main package
       silo = craneLib.buildPackage (commonArgs // {
         inherit cargoArtifacts;
         doCheck = false; # Tests require external services, CI runs them separately
       });
+
+      # Debug build (fast to compile, no optimizations)
+      debugArgs = {
```
airhorns, line 60: :metal

airhorns (approved):
:LGTM:! Would you mind adding a quick thing to the docs to describe this property? 

And also, I think we might wanna talk right to silo from the sandbox in the future for workflow things where we latch step progress as we go, which will require a more sophisticated auth setup at that time where we need signed tokens that are only allowed to transact against one tenant, but we can cross that bridge when we come to it.

PR #140, opened 2026-02-20
Add deploy-local-orbstack script that deploys .#silo-docker-dev package

kirinrastogi:
## Summary

Adds `scripts/deploy-local-orbstack.mts`, a script to build and deploy silo to a local OrbStack Kubernetes cluster.

You may need to run `kubectl --context orbstack delete namespace silo-test` first for a fresh start.

I wasn't able to get cross compilation working, so it builds inside of orbstack for an aarch64-linux artifact. The first time this is run it will take ~15 minutes, and the cache persists within orbstack. Subsequent runs are fast (~3s total)

### Building inside orbstack

OrbStack runs Linux containers natively on Apple Silicon, which means Kubernetes pods need `aarch64-linux` binaries. However, running `nix build .#silo-docker-dev` on a Mac produces an `aarch64-darwin` artifact — the wrong OS entirely.

Rather than cross-compiling, the script runs the Nix build *inside* OrbStack itself using a `nixos/nix` container, which is natively `aarch64-linux`. The resulting Docker image tarball is piped directly to `docker load`, then the manifests are applied and the StatefulSet is restarted to pick up the new image.

A named Docker volume (`silo-nix-store`) caches the Nix store between runs so subsequent builds are fast.


airhorns:
Ah -- this dockerfile is really just janky code for local testing, I think we should ideally instead use the nix-built docker container if we can. We use this fancy feature of nix that produces minimal docker images from your nix config without needing a dockerfile: https://github.com/gadget-inc/silo/blob/main/nix/modules/docker.nix

That's what CI is building and that's what we'd ship to prod, but I added this silly dockerfile at one point cause I hit some issues getting that working. I think that is now working well though and we should probably just delete it, and make the local dev script use the nix-built one instead, as that should always be working! Want to do that instead?

PR #158, opened 2026-03-05
Switch tenant sharding from lexicographic ranges to hash-based

kirinrastogi:
## Summary

Tenant IDs are now hashed before shard assignment, preventing hotspots where tenants with shared prefixes (e.g. all `env-*` tenants) cluster on a single shard.

- **Hash algorithm**: XXH64 via `xxhash-rust` (Rust) and `xxhash-wasm` (TypeScript) — a well-known, fast non-cryptographic hash with excellent avalanche properties. Using crate/npm packages on both sides means we don't maintain our own hash implementation.
- **Raw keys preserved**: The tenant key itself is still stored everywhere (shard databases, logs, gRPC requests). Hashing only happens at routing time to decide which shard owns a tenant. This means a marginal increase in CPU for hashing on each request, but we get to see raw tenant keys in logs and storage.
- **Range boundaries**: Shard ranges are now evenly-spaced 16-char hex strings in the u64 hash space. Midpoint computation is numeric (u64 arithmetic) rather than the old string-based approach.

## Test plan
- [x] All shard range, split, and coordination tests pass with hash-space boundaries
- [x] Cross-language hash parity verified (Rust and TypeScript produce identical XXH64 outputs)
- [x] etcd and k8s sequential split tests use offset split points (not exact midpoints)
- [x] Determinism test verifies same tenant always maps to same shard

### DIFF
--- a/src/shard_range.rs
+++ b/src/shard_range.rs
@@ -198,9 +198,9 @@ impl ShardRange {
     /// Compute the numeric midpoint of this hash-space range for auto-split.
     ///
     /// Interprets the range boundaries as 16-character hex-encoded u64 values
-    /// and returns the midpoint as a 16-character hex string.
+    /// and returns the midpoint as a u64.
     /// Returns None if the range is too small to split.
-    pub fn midpoint(&self) -> Option<String> {
+    pub fn midpoint(&self) -> Option<u64> {
         let start_val: u64 = if self.start.is_empty() {
             0
         } else {
@@ -217,8 +217,7 @@ impl ShardRange {
             return None;
         }
 
-        let mid = start_val + (end_val - start_val) / 2;
-        Some(format!("{:016x}", mid))
+        Some(start_val + (end_val - start_val) / 2)
     }
 }
 
@@ -419,22 +418,20 @@ impl ShardMap {
             // Single shard covers entire keyspace
             shards.push(ShardInfo::new(ShardId::new(), ShardRange::full()));
         } else {
-            // Divide keyspace into equal ranges
-            // We use a simple approach: divide into ranges based on first character
-            // For better distribution, we create ranges using hex boundaries
+            // Divide the u64 hash keyspace into equal ranges
             let boundaries = compute_range_boundaries(shard_count);
 
             for i in 0..shard_count as usize {
                 let start = if i == 0 {
                     String::new() // Unbounded start
                 } else {
-                    boundaries[i - 1].clone()
+                    format_hash_boundary(boundaries[i - 1])
                 };
 
                 let end = if i == shard_count as usize - 1 {
                     String::new() // Unbounded end
                 } else {
-                    boundaries[i].clone()
+                    format_hash_boundary(boundaries[i])
                 };
 
                 shards.push(ShardInfo::new(ShardId::new(), ShardRange::new(start, end)));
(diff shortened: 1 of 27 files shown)


src/shard_range.rs
```diff
@@ -711,40 +678,19 @@ pub enum ShardMapError {
...
-            let boundary = (i * 256 / shard_count) as u8;
-            boundaries.push(format!("{:02x}", boundary));
-        }
-    } else {
-        // Use four hex digits for larger shard counts
-        for i in 1..shard_count {
-            let boundary = (i as u64 * 65536 / shard_count as u64) as u16;
-            boundaries.push(format!("{:04x}", boundary));
-        }
+    for i in 1..shard_count {
```
kirinrastogi, line 693: this is possible because the values are now always 16 char hex values.

src/shard_range.rs
```diff
@@ -177,81 +195,30 @@ impl ShardRange {
...
-        let end_char = end_bytes[diff_pos];
-
-        if end_char <= start_char + 1 {
-            // Too close to split at this level, extend
-            let mut mid = String::from_utf8_lossy(&start_bytes[..=diff_pos]).to_string();
-            mid.push('~'); // High ASCII value
-            if mid.as_str() > effective_start.as_str() && mid.as_str() < effective_end.as_str() {
-                return Some(mid);
-            }
+        if end_val <= start_val + 1 {
```
kirinrastogi, line 216: related test added here: https://github.com/gadget-inc/silo/pull/158/changes#diff-38a9073e91f709944f8a3130d1558a8143e5ddca03557050d4c2b7fd9a00c2b7R28

src/shard_range.rs
```diff
@@ -177,81 +195,30 @@ impl ShardRange {
...
-        let mut result = String::from_utf8_lossy(&start_bytes[..diff_pos]).to_string();
-        result.push(mid_char as char);
-
-        // Validate the result
-        if result.as_str() > effective_start.as_str() && result.as_str() < effective_end.as_str() {
-            Some(result)
-        } else {
-            None
-        }
+        let mid = start_val + (end_val - start_val) / 2;
```
kirinrastogi: this is simplified now because of fixed length (16 char hex) boundary values

typescript_client/src/client.ts
```diff
@@ -1358,6 +1390,7 @@ export class SiloGRPCClient {
    * @internal
    */
   private async _ensureTopology(): Promise<void> {
+    await initHasher();
```
airhorns, line 1393: 🙄 👍

src/shard_range.rs
```diff
@@ -177,81 +195,30 @@ impl ShardRange {
...
     }
 
-    /// Compute the lexicographic midpoint of this range for auto-split.
+    /// Compute the numeric midpoint of this hash-space range for auto-split.
     ///
-    /// Returns None if the range cannot be split (e.g., single character range).
+    /// Interprets the range boundaries as 16-character hex-encoded u64 values
+    /// and returns the midpoint as a 16-character hex string.
+    /// Returns None if the range is too small to split.
     pub fn midpoint(&self) -> Option<String> {
```
airhorns: should we just change the signature to return this as the numeric midpoint? since thats what the ranges really are, may as well avoid stringifying and parsing?

src/shard_range.rs
```diff
@@ -711,40 +678,19 @@ pub enum ShardMapError {
 
 /// Compute evenly-spaced range boundaries for the given shard count.
 ///
-/// Returns `shard_count - 1` boundary strings that divide the keyspace.
-/// Uses a simple hex-based scheme for predictable distribution.
+/// Returns `shard_count - 1` boundary strings that divide the u64 hash
+/// keyspace into equal-sized partitions, formatted as 16-character hex strings.
 fn compute_range_boundaries(shard_count: u32) -> Vec<String> {
```
airhorns: same question

airhorns (approved):
LGTM after comments! Nice and simple and even deletes implementation code :)

airhorns:
Oh also would be worth tasking a claude to see if any docs updates are necessary

kirinrastogi:
going to merge on green

PR #165, opened 2026-03-10
Fix siloctl to use xxhash-based routing for shard splitting

kirinrastogi:
## Summary
- `shard split --auto` was computing midpoints using a lexicographic string algorithm with incorrect bounds (`"zzzzzzzz"`), but shard ranges are now 16-char hex-encoded u64 hash values from xxhash. Replaced with delegation to `ShardRange::midpoint()` and `format_hash_boundary()`, matching the hash-space routing used by the rest of the system
- Fix `siloctl cluster info` column widths — the Range column was 14 chars wide but hash-based ranges like `[2000000000000000, 4000000000000000)` are 40 chars, causing misaligned output. Widened to 40
- Add `siloctl tenant hash <tenant_id>` command to compute the xxhash for a tenant ID without needing a server connection. Supports `--json` output

## Test plan
- [x] All 25 siloctl tests pass
- [x] `siloctl cluster info` output aligns correctly with hash-based ranges
- [x] `siloctl tenant hash` returns correct xxhash values matching the routing client
- [x] Verify `siloctl shard split --auto` produces correct hex midpoints against a running cluster

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5377,7 +5377,7 @@ dependencies = [
 
 [[package]]
 name = "silo"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "anyhow",
  "askama",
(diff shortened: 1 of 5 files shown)


airhorns (approved):
Dope 👍

PR #181, opened 2026-03-17
Return failed_precondition when importing a running job

kirinrastogi:
## Summary

- ImportJobs RPC now returns a `failed_precondition` gRPC error when attempting to import a job that already exists and is currently in the Running state
- Previously this was returned as a per-job `success=false` result; now it surfaces as a proper RPC-level error
- Added `JobNotReimportable` handling to the `map_err` function in the server

## Context

For importing jobs we are doing a drain-style cutover: jobs must complete their running attempts before being imported to Silo. If they have extra attempts that's fine, as long as they are not currently running. This change makes the constraint explicit at the RPC level so callers get a clear error if they try to import a job that hasn't finished draining yet.

## Test plan

- [x] Updated `reimport_running_job_fails` to verify `import_jobs` returns `Err` (not a per-job result) when the target job is Running
- [x] Added `reimport_running_job_in_batch_fails_entire_call` to verify a batch import fails entirely when any job in the batch is Running
- [x] All 80 import tests pass

### DIFF
--- a/src/job_store_shard/import.rs
+++ b/src/job_store_shard/import.rs
@@ -87,11 +87,27 @@ pub struct ImportJobParams {
 impl JobStoreShard {
     /// Import a batch of jobs, returning per-job results.
     /// Each job is imported independently; failures don't affect other jobs in the batch.
+    /// Returns an error if any job in the batch already exists and is currently Running.
     pub async fn import_jobs(
         &self,
         tenant: &str,
         jobs: Vec<ImportJobParams>,
     ) -> Result<Vec<ImportJobResult>, JobStoreShardError> {
+        // Pre-check: reject the entire batch if any job is currently Running.
+        for params in &jobs {
+            if let Some(status) = self.get_job_status(tenant, &params.id).await? {
+                if status.kind == JobStatusKind::Running {
+                    return Err(JobStoreShardError::JobNotReimportable(
+                        JobNotReimportableError {
+                            job_id: params.id.clone(),
+                            status: JobStatusKind::Running,
+                            reason: "job is currently running".to_string(),
+                        },
+                    ));
+                }
+            }
+        }
+
         let mut results = Vec::with_capacity(jobs.len());
         for params in jobs {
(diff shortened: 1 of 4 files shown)


tests/job_store_shard_import_tests.rs
```diff
@@ -1334,18 +1334,50 @@ async fn reimport_running_job_fails() {
...
+    reimport_b.attempts = vec![failed_attempt(1_700_000_001_000)];
+
+    let err = shard
+        .import_jobs("-", vec![reimport_a, reimport_b])
+        .await
+        .expect_err("batch import should fail when any job is running");
+    let err_msg = err.to_string();
+    assert!(
+        err_msg.contains("currently running"),
+        "error should mention 'currently running', got: {err_msg}"
```
airhorns, line 1381: can you expand this to test if the one job that was not running was successfully imported? i think that it is because we dont do huge import transactions, and thats why the original api returned a per-job status. i dont feel strongly it needs to be that way, i just think it happens to be that way right now

kirinrastogi: I updated it so it will scan the entire batch before starting to commit.

that will prevent partial applies, and fits better with a failed precondition

src/job_store_shard/import.rs
```diff
@@ -87,11 +87,27 @@ pub struct ImportJobParams {
 impl JobStoreShard {
     /// Import a batch of jobs, returning per-job results.
     /// Each job is imported independently; failures don't affect other jobs in the batch.
+    /// Returns an error if any job in the batch already exists and is currently Running.
     pub async fn import_jobs(
         &self,
         tenant: &str,
         jobs: Vec<ImportJobParams>,
     ) -> Result<Vec<ImportJobResult>, JobStoreShardError> {
+        // Pre-check: reject the entire batch if any job is currently Running.
```
airhorns, line 96: 👍

airhorns (approved):

PR #209, opened 2026-03-22
Add in-memory lease tracker to eliminate DB scans during reaping

airhorns:
## Summary
- Add `LeaseTracker` backed by `DashMap<String, i64>` (task_id → expiry_ms) to track active leases in memory
- Hydrated once from DB on shard startup, then kept in sync via insert/remove/update calls from dequeue, heartbeat, and report_outcome
- `reap_expired_leases` now iterates the in-memory map (~microseconds) and only does point reads for expired leases, instead of scanning all lease keys in SlateDB
- Increase reaper interval from 250ms to 1s

## Context
Production profiling showed lease reaping consuming ~14% of CPU despite only 10s-100s of active leases. The cost came from SlateDB's merge iterator setup on every scan (SST index parsing, block reads) and iterating through tombstones from deleted leases. Benchmarking confirmed p50 reap latency degraded from 1.1ms to 48ms after 50K job churn cycles.

Memory cost is ~300 bytes per active lease — negligible even at 10K concurrent leases.

## Test plan
- [x] `cargo check` passes
- [ ] CI tests pass
- [ ] Verify in production that lease reaper CPU drops significantly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes core lease lifecycle and expired-lease reaping to rely on a new in-memory index; any missed insert/remove/update or hydration edge case could leave leases unreaped or incorrectly reaped.
> 
> **Overview**
> **Eliminates full-DB scans during lease reaping** by introducing an in-memory `LeaseManager` that hydrates once from existing lease keys on shard open, then stays in sync via updates from `dequeue`, `lease_task`, `heartbeat_task`, `report_attempt_outcome`, and floating-limit refresh reporting.
> 
> `reap_expired_leases` is refactored to iterate the tracked in-memory leases and only perform DB point reads when needed (notably for `RefreshFloatingLimit` cleanup), reducing tombstone/iterator overhead.
> 
> Adds a new `scan_tombstone_profile` benchmark plus targeted tests/utilities to force-expire leases and validate DB↔manager consistency and crash rehydration behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e02bf2e593b936ee1721aa32b22dd0a9b0d57264. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/job_store_shard/lease.rs
+++ b/src/job_store_shard/lease.rs
@@ -368,11 +368,6 @@ impl JobStoreShard {
     /// Uses the in-memory lease tracker instead of scanning the DB.
     /// Returns the number of expired leases reaped.
     pub async fn reap_expired_leases(&self, _tenant: &str) -> Result<usize, JobStoreShardError> {
-        // Wait for lease tracker to be hydrated from DB before reaping
-        if !self.lease_tracker.is_hydrated() {
-            return Ok(0);
-        }
-
         let now_ms = now_epoch_ms();
         let expired = self.lease_tracker.expired_leases(now_ms);
 
@@ -385,6 +380,7 @@ impl JobStoreShard {
         for lease in &expired {
             let task_id = &lease.task_id;
 
+            // [SILO-REAP-1] Pre: Lease exists (found in in-memory tracker)
             // Read the lease from DB to get full details needed for reaping
             let key = leased_task_key(task_id);
             let maybe_raw = self.db.get(&key).await?;
@@ -402,6 +398,7 @@ impl JobStoreShard {
                 }
             };
 
+            // [SILO-REAP-2] Pre: Check if lease has expired
             // Double-check expiry (could have been renewed via heartbeat since we read the tracker)
             if decoded.expiry_ms() > now_ms {
                 continue;
@@ -428,6 +425,9 @@ impl JobStoreShard {
             {
                 AttemptOutcome::Cancelled
             } else {
+                // [SILO-REAP-3][SILO-REAP-4] Report as worker crashed
+                // SILO-REAP-3: Post: Set job status to Failed (via report_attempt_outcome)
+                // SILO-REAP-4: Post: Set attempt status to AttemptFailed
                 AttemptOutcome::Error {
                     error_code: "WORKER_CRASHED".to_string(),
                     error: format!(
@@ -440,7 +440,7 @@ impl JobStoreShard {
                 }
             };
 
-            // report_attempt_outcome removes the lease from the tracker via its own remove() call
+            // [SILO-REAP-REL] Release lease and update job/attempt status via report_attempt_outcome
             let _ = self.report_attempt_outcome(task_id, outcome).await;
             reaped += 1;
         }
--- a/src/job_store_shard/lease.rs
+++ b/src/job_store_shard/lease.rs
@@ -368,6 +368,12 @@ impl JobStoreShard {
     /// Uses the in-memory lease tracker instead of scanning the DB.
     /// Returns the number of expired leases reaped.
     pub async fn reap_expired_leases(&self, _tenant: &str) -> Result<usize, JobStoreShardError> {
+        if !self.lease_tracker.is_hydrated() {
+            self.lease_tracker
+                .hydrate_from_db(self.db.as_ref(), &self.range)
+                .await;
+        }
+
         let now_ms = now_epoch_ms();
         let expired = self.lease_tracker.expired_leases(now_ms);
 
@@ -407,9 +413,19 @@ impl JobStoreShard {
             // Handle RefreshFloatingLimit tasks separately
             if let Some((task_id, queue_key)) = decoded.refresh_floating_limit_info() {
                 let task_tenant = decoded.tenant();
-                let _ = self
+                if let Err(error) = self
                     .reap_expired_refresh_task(task_tenant, task_id, queue_key, &decoded)
-                    .await;
+                    .await
+                {
+                    tracing::warn!(
+                        tenant = %task_tenant,
+                        task_id = %task_id,
+                        queue_key = %queue_key,
+                        error = %error,
+                        "failed to reap expired floating limit refresh lease"
+                    );
+                    continue;
+                }
                 self.lease_tracker.remove(&lease.task_id);
                 reaped += 1;
                 continue;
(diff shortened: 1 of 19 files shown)


src/job_store_shard/lease.rs
```diff
@@ -357,76 +365,82 @@ impl JobStoreShard {
...
-        let shard_range = self.get_range();
+        for lease in &expired {
+            let task_id = &lease.task_id;
 
-        // Collect defunct lease keys to delete
-        let mut defunct_keys: Vec<Vec<u8>> = Vec::new();
+            // [SILO-REAP-1] Pre: Lease exists (found in in-memory tracker)
+            // Read the lease from DB to get full details needed for reaping
+            let key = leased_task_key(task_id);
+            let maybe_raw = self.db.get(&key).await?;
```
airhorns: this get seems unnecessary and slow, can we just blindly delete from the db regardless?

src/job_store_shard/lease.rs
```diff
@@ -526,3 +524,109 @@ impl JobStoreShard {
         Ok(())
     }
 }
+
+#[cfg(test)]
```
airhorns: we dont put tests inline in rust source files, they must go in a file in tests/ instead, either create a new file or add these to an existing file

src/job_store_shard/lease.rs
```diff
@@ -526,3 +524,109 @@ impl JobStoreShard {
...
+
+    use super::*;
+    use crate::codec::{decode_lease, encode_lease};
+    use crate::gubernator::MockGubernatorClient;
+    use crate::job::JobStatusKind;
+    use crate::settings::{Backend, DatabaseConfig};
+    use crate::shard_range::ShardRange;
+    use crate::task::LeaseRecord;
+
+    async fn open_temp_shard() -> (tempfile::TempDir, Arc<JobStoreShard>) {
```
airhorns: we have test helpers for this elsewhere already

src/job_store_shard/lease_task.rs
```diff
@@ -217,6 +217,10 @@ impl JobStoreShard {
         })
         .await?;
 
+        // Keep the in-memory tracker aligned with leases created outside dequeue.
```
airhorns:
```suggestion
        // Record the lease in the lease manager after durably commiting it
```

src/job_store_shard/mod.rs
```diff
@@ -107,6 +108,8 @@ pub struct JobStoreShard {
     db_path: String,
     /// The shard's tenant range, immutable after opening.
     range: ShardRange,
+    /// In-memory lease tracker to avoid expensive DB scans during reaping.
```
airhorns:
```suggestion
    /// In-memory lease tracker that avoids expensive DB scans to check lease expiry.
```

src/lease_manager.rs
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    /// Remove a lease (called when a lease is released via report_outcome or reaping).
+    pub fn remove(&self, task_id: &str) {
+        self.leases.remove(task_id);
+    }
+
+    /// Returns true if the initial DB hydration scan has completed.
+    pub fn is_hydrated(&self) -> bool {
+        self.hydrated.load(Ordering::Acquire)
```
airhorns: i think we should make lease hydrating an internal concept to the lease manager, where it internally awaits hydration for responding to any methods that need it to complete ahead of time. all the callers currently are checking if its hydrated, which makes it a leaky abstraction. we should keep the eager async hydration kickoff on shard open, but then just make the interface "work" after that by having the lease manager await hydration internally before returning from functions that require it

src/lease_manager.rs
```diff
@@ -0,0 +1,205 @@
...
+        if self.is_hydrated() {
+            return;
+        }
+
+        if self
+            .hydrating
+            .compare_exchange(false, true, Ordering::AcqRel, Ordering::Acquire)
+            .is_err()
+        {
+            // Another task is already hydrating — wait for it to finish.
```
airhorns: this is weird -- we should never be hydrating twice at the same time, design that out of existence somehow

src/job_store_shard/lease.rs
```diff
@@ -357,76 +365,76 @@ impl JobStoreShard {
...
-        let mut defunct_keys: Vec<Vec<u8>> = Vec::new();
+            // [SILO-REAP-1] Pre: Lease exists (found in in-memory tracker)
+            // Read the lease from DB to get full details needed for reaping
+            let key = leased_task_key(task_id);
+            let maybe_raw = self.db.get(&key).await?;
+            let Some(value_bytes) = maybe_raw else {
+                // Lease was already removed (e.g., worker completed between check and reap)
+                self.lease_manager.remove(task_id);
+                continue;
+            };
```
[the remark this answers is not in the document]
airhorns: Resolved in the latest commit — RunAttempt leases no longer do any db.get during reaping. The tracker stores all needed metadata.

src/job_store_shard/lease.rs
```diff
@@ -357,76 +365,76 @@ impl JobStoreShard {
...
-                    range = %shard_range,
-                    "deleting defunct lease (tenant outside shard range)"
-                );
-                defunct_keys.push(kv.key.to_vec());
-                continue;
-            }
-
-            // [SILO-REAP-1] Pre: Lease exists (we found it)
             // [SILO-REAP-2] Pre: Check if lease has expired
+            // Double-check expiry (could have been renewed via heartbeat since we read the tracker)
```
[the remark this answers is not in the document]
airhorns: No longer applicable — the latest commit removed the double-check against DB expiry for RunAttempt leases. The tracker is authoritative since all lease mutations (dequeue, heartbeat, report_outcome) go through the same process.

src/job_store_shard/lease.rs
```diff
@@ -449,22 +457,6 @@ impl JobStoreShard {
             reaped += 1;
         }
 
-        // Delete defunct leases from the database
-        if !defunct_keys.is_empty() {
-            let mut batch = WriteBatch::new();
-            for key in &defunct_keys {
-                batch.delete(key);
-            }
```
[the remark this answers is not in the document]
airhorns: Pre-existing behavior — the `let _ =` pattern was in the original reaper code before the lease manager changes. The reaper is intentionally best-effort: if report_attempt_outcome fails, the lease remains in both DB and tracker, so the next reap cycle will retry. The count is only used for metrics/logging, not correctness.

PR #221, opened 2026-03-30
Add SiloAutoscaler CRD and controller for safe StatefulSet scaling

kirinrastogi:
## Summary
- New `silo-autoscaler` subcrate: Kubernetes controller that manages safe scaling of Silo StatefulSets using pod finalizers
- CRD with `/scale` subresource for HPA targeting
- Finalizer-based scale-down: gates pod deletion until shard leases are released, automatically recovers from SIGKILL'd pods
- ValidatingAdmissionPolicy with CEL for spec validation
- Integration tests in dedicated `silo-autoscaler-test` namespace

## Tophat

Deploy silo and the autoscaler to your local OrbStack cluster:

```sh
./scripts/deploy-local-orbstack.mts
```

Then create or edit the SiloAutoscaler to change replicas:

```sh
kubectl --context orbstack -n silo-test apply -f - <<EOF
apiVersion: silo.dev/v1alpha1
kind: SiloAutoscaler
metadata:
  name: silo-autoscaler
  namespace: silo-test
spec:
  replicas: 2
  targetStatefulSet: silo
  clusterPrefix: silo-local-test
EOF
```

Watch the pods scale:

```sh
kubectl --context orbstack -n silo-test get pods -l app=silo -w
```

```sh
kubectl --context orbstack get siloautoscaler -n silo-test
```

## Test plan
- [x] Unit tests: `cargo test --package silo-autoscaler --lib` (10 tests)
- [x] Integration tests: `cargo test --package silo-autoscaler --test autoscaler_integration_tests -- --test-threads=1` (3 tests)
- [x] Manual testing on OrbStack: scale up, scale down, orphaned lease recovery


silo-autoscaler/src/controller.rs
```diff
@@ -0,0 +1,529 @@
...
+            TerminationStatus::Recovery { orphaned_leases } => orphaned_leases,
+            _ => vec![],
+        }
+    }
+}
+
+/// Run the SiloAutoscaler controller loop.
+pub async fn run_controller(client: Client) -> anyhow::Result<()> {
+    let autoscalers: Api<SiloAutoscaler> = Api::all(client.clone());
+    let pods: Api<Pod> = Api::all(client.clone());
```
udnay: I assume there is a selector on this somehow? That `Api::all(client.clone())` is not actually all pods.

udnay (approved):
This looks good to me. Can you create a gameday or update the runbook on how we should handle the loss of a PVC that's not at the end or the stateful set list? If we have `silo-0`, `silo-1`, `silo-2` and `silo-1` has a pvc failure what do we do? Scale down to `1` and then back to `3`?

kirinrastogi:
@udnay I will do that before enabling this in production. I want to get it deployed in staging first. 

I think for any pvc failure we would either wait for the PVC to be fixed or accept data loss and use siloctl's force-release command.

silo-autoscaler/src/controller.rs
kirinrastogi: I updated this to check the namespace that the controller is in instead.

kirinrastogi:
Going to merge. We could expand this in the future to support statefulsets of size 1. Not a trapdoor decision to deploy it as is.

PR #225, opened 2026-04-07
Drop completed jobs older than 7 days via slatedb compaction filter

kirinrastogi:
## Summary

Adds a `CompactionFilter` (`CompletedJobCompactionFilter`) that tombstones data for completed jobs older than a 7-day retention window during slatedb compaction. The filter handles multiple key prefixes in a single compaction pass:

- **JOB_STATUS (0x02):** Decodes the value; if the status is terminal (Succeeded/Failed/Cancelled) and `changed_at_ms` is older than the 7-day cutoff, tombstones the row and records the `job_id` in an in-memory `HashSet`.
- **IDX_STATUS_TIME (0x03):** Self-contained — the key encodes the status string and timestamp, so the filter decides without a value decode or set lookup.
- **IDX_METADATA (0x04), ATTEMPT (0x07), JOB_CANCELLED (0x0A):** The key encodes `job_id`. The filter checks membership in the `HashSet` built from JOB_STATUS rows (which sort before these prefixes).

Tombstone (rather than `Drop`) is used so older non-terminal versions in lower sorted runs are shadowed and can't resurrect.

The 7-day clock starts when the job transitions to a terminal state (`changed_at_ms` on the terminal `JobStatus`), not when it was originally enqueued.

### JOB_INFO (0x01) is not handled

`JOB_INFO` rows use prefix 0x01, which sorts *before* `JOB_STATUS` (0x02). During a single-pass compaction, the filter sees all JOB_INFO rows before any JOB_STATUS rows, so it can't know which jobs are completed yet. Cleaning up JOB_INFO would require either a separate scan-based cleanup pass or a distributed/two-pass compaction model — this might be better suited for a future slatedb update that supports distributed compaction, where a pre-scan phase could identify eligible job_ids before the main compaction pass runs.

### Wiring

The filter is installed by constructing a custom `slatedb::CompactorBuilder` in `JobStoreShard::open_with_resolved_store`. Any user-supplied `compactor_options` from the slatedb settings are forwarded onto the builder so existing tunables still apply. An optional `compaction_filter_supplier` field on `OpenShardOptions` allows benches/tests to override the retention window (production callers pass `None` and get the 7-day default).

Also includes a `compaction_filter_impact` benchmark with two scenarios:
- **Scenario A:** Steady-state with ongoing background compaction (configurable duration, default 5 min).
- **Scenario B:** Latency before vs after one manual compaction (configurable job count, default 50k).

## Test plan
- [x] 11 unit tests covering drop/keep for all handled prefixes (`cargo test --lib job_store_shard::compaction_filter`).
- [x] 2 integration tests: one verifies all related rows are dropped for old completed jobs; the other verifies all rows survive for recently completed jobs (`cargo test --test compaction_filter_tests`).
- [x] `cargo check --tests` and `cargo check --benches` pass.
- [x] Benchmark validated at 90s/5k and 300s/50k scales — filter adds no measurable throughput regression under ongoing compaction; post-compaction read p99 improves ~30% vs baseline due to smaller live set.


kirinrastogi:
we need a standalone approach that handles metadata, secondary indexes, and counter/merge updates

PR #236, opened 2026-04-14
Cache split-paused state locally so is_shard_paused never hits the k8s API

udnay:
## Summary
  - Add `paused_shards` local cache to `CoordinatorBase` so `is_shard_paused` is a lock-free read instead of a k8s/etcd API call
  - Update split execution, abandon, and completion paths to keep the cache in sync
  - On split resumption, re-hydrate the cache from persisted split state
  - Add `CoordinatorBase::new_with_shared` constructor for tests that share `shard_map`/`owned` Arcs
  - Add server-level test: enqueue returns UNAVAILABLE on a paused shard and succeeds on a healthy shard, even when the k8s API is completely unreachable

---

> [!NOTE]
> **Medium Risk**
> Changes shard request routing behavior by making `Coordinator::is_shard_paused` rely on an in-memory cache that must stay consistent with persisted split records; bugs could cause shards to incorrectly accept or reject traffic during/after splits and restarts.
> 
> **Overview**
> **Avoids k8s/etcd reads on the hot path** by introducing a `CoordinatorBase` in-memory `paused_shards` cache and switching the default `Coordinator::is_shard_paused` implementation to consult it instead of loading split state.
> 
> **Keeps the cache in sync with split lifecycle**: split execution now marks shards paused/unpaused as phases change, clears paused state when abandoning or cleaning up splits, and adds `ShardSplitter::rehydrate_paused_shards` to repopulate paused state from persisted split records after restarts (called during startup in `main`).
> 
> **Tests updated/added** to validate restart rehydration, stale-split recovery clearing the cache, and server behavior when the coordination backend is unreachable (paused shard returns UNAVAILABLE while healthy shards still work).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit edab0bad14d109307178cd04f365ae18669b7552. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/coordination/mod.rs
+++ b/src/coordination/mod.rs
@@ -367,6 +367,11 @@ pub struct CoordinatorBase {
     /// Placement rings this node participates in.
     /// Empty means the node participates in the default ring only.
     pub placement_rings: Vec<String>,
+    /// Cache of shards that are currently paused for split operations.
+    /// Updated by the split execution code when phases change.
+    /// Read by `is_shard_paused` on the hot request path to avoid k8s API calls.
+    /// Uses std::sync::RwLock (not tokio) because the critical section is tiny.
+    paused_shards: Arc<std::sync::RwLock<HashSet<ShardId>>>,
 }
 
 impl CoordinatorBase {
@@ -393,6 +398,31 @@ impl CoordinatorBase {
             factory,
             startup_time_ms,
             placement_rings,
+            paused_shards: Arc::new(std::sync::RwLock::new(HashSet::new())),
+        }
+    }
+
+    /// Create a coordinator base with pre-built shared state.
+    /// Used by tests that need to share `shard_map` and `owned` Arcs with mock coordinators.
+    pub fn new_with_shared(
+        node_id: impl Into<String>,
+        grpc_addr: impl Into<String>,
+        shard_map: Arc<Mutex<ShardMap>>,
+        owned: Arc<Mutex<HashSet<ShardId>>>,
+        factory: Arc<ShardFactory>,
+    ) -> Self {
+        let (shutdown_tx, shutdown_rx) = watch::channel(false);
+        Self {
+            node_id: node_id.into(),
+            grpc_addr: grpc_addr.into(),
+            shard_map,
+            owned,
+            shutdown_tx,
+            shutdown_rx,
+            factory,
+            startup_time_ms: None,
+            placement_rings: Vec::new(),
+            paused_shards: Arc::new(std::sync::RwLock::new(HashSet::new())),
         }
     }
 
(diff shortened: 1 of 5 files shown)


udnay:

- #236 ◀

src/main.rs
```diff
@@ -117,6 +117,22 @@ async fn main() -> anyhow::Result<()> {
             }
         };
 
+    // Re-hydrate the paused-shards cache from any persisted split records
+    // left behind by a previous incarnation. This ensures is_shard_paused
+    // correctly rejects traffic to shards that were mid-split before the
+    // process restarted. The split records themselves are NOT cleaned up —
```
udnay, line 123: This was my decision to not clean up the splits right away, I don't know if it's safe, claude seems to think so but I wonder if we do want an operator in the middle seeing as how we don't call `recover_stale_splits` all the time at startup.

src/main.rs
```diff
@@ -117,6 +117,22 @@ async fn main() -> anyhow::Result<()> {
...
+    // an operator must decide whether to abandon them (e.g. via siloctl or
+    // recover_stale_splits).
+    {
+        let splitter = silo::coordination::ShardSplitter::new(coordinator.clone());
+        splitter.rehydrate_paused_shards().await.map_err(|e| {
+            anyhow::anyhow!(
+                "failed to re-hydrate paused shards from split records: {}",
```
udnay, line 130: Fail hard here, if we can't talk to k8s api then we should not keep going. I'm not sure how we would be able to restart a pod if the control plane is down but better safe than not.

src/coordination/mod.rs
```diff
@@ -393,6 +398,31 @@ impl CoordinatorBase {
             factory,
             startup_time_ms,
             placement_rings,
+            paused_shards: Arc::new(std::sync::RwLock::new(HashSet::new())),
+        }
+    }
+
+    /// Create a coordinator base with pre-built shared state.
+    /// Used by tests that need to share `shard_map` and `owned` Arcs with mock coordinators.
+    pub fn new_with_shared(
```
kirinrastogi, line 408: nit: add `#[cfg(test)]` so this is only included in test code

src/coordination/split.rs
```diff
@@ -772,6 +787,39 @@ impl ShardSplitter {
...
+    /// traffic-pausing phase so that `is_shard_paused` correctly rejects
+    /// requests until an operator decides what to do with the stale split.
+    ///
+    /// This does NOT delete or modify split records — call
+    /// `recover_stale_splits` (or use siloctl) if you want to abandon them.
+    pub async fn rehydrate_paused_shards(&self) -> Result<(), CoordinationError> {
+        let splits = self.coordinator.list_all_splits().await?;
+
+        for split in splits {
+            if split.initiator_node_id != self.ctx.node_id {
```
kirinrastogi, line 804: 👍🏽

kirinrastogi (approved):
Looks good. I think we need to do something similar for `GetClusterInfo` next

src/coordination/mod.rs
udnay: Done

PR #239, opened 2026-04-17
Relax SILO-COORD-INV-4: partitioned nodes keep serving

kirinrastogi:
While running the [node-flap gameday](https://github.com/gadget-inc/global-infrastructure/pull/1736) I noticed that the gameday says that a node that cannot reach the k8s api for leases should not be able to serve requests, but that is wrong, and silo allows traffic during this partition.

Shard leases are permanent, and that is what allows the node to continue serving safely during this partition. Even though the membership lease expires, no other silo nodes will acquire the shard and have a split brain (as the shard lease cannot be and is not deleted during this partition).

## Summary
- Restate SILO-COORD-INV-4: a node partitioned from the coordination layer can continue to serve requests for shards it already owns, but cannot acquire or release shard leases while partitioned (mutations require current membership).
- Update the Alloy model: `canServe` now only requires `holdsLease`; `leaseReleaseStep` gains `isMember` / `not hasFlapped` preconditions (acquire already had them); the former `flappedNodeCannotServe` assertion is replaced with `flappedNodeCannotMutateLeases`, which asserts no `leaseAcquireStep` or `leaseReleaseStep` can be taken while flapped.
- Align example traces (`exampleNodeFlapPermanentLease`, `exampleNodeRestart`) with the new serving semantics.

## Test plan
- [x] `alloy6 exec -f -o specs/output specs/coordination.als` — all 16 `check`s UNSAT (no counterexamples), all 9 `run` examples SAT.

### DIFF
--- a/specs/coordination.als
+++ b/specs/coordination.als
@@ -455,16 +455,15 @@ pred hasLeaseHolder[s: Shard, t: Time] {
 
 /**
  * Check if a node can safely serve a shard.
- * 
- * A node can serve a shard only if:
- * - It holds the lease for the shard
- * - It has not flapped (lost connectivity)
- * - It is still a member (from etcd's perspective)
+ *
+ * A node can serve a shard whenever it holds the (permanent) lease. In
+ * particular, a node that is partitioned from the coordination layer (flapped)
+ * continues to serve traffic for shards it already owns — holding the lease is
+ * sufficient authority to serve. Lease *mutation* (acquire/release) still
+ * requires current membership; that is enforced at the step level.
  */
 pred canServe[n: Node, s: Shard, t: Time] {
     holdsLease[n, s, t]
-    isMember[n, t]
-    not hasFlapped[n, t]
 }
 
 /**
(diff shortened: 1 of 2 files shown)


airhorns (approved):
Right, this makes sense, I missed this divergence when switching from the actively-held lease approach to the perma-lease approach. Thanks for fixing!

specs/coordination.als
```diff
@@ -1589,14 +1588,20 @@ pred leaseAcquireStep[n: Node, s: Shard, t: Time, tnext: Time] {
...
+ * [SILO-COORD-INV-4] A partitioned node cannot release its leases: mutation
+ * requires current membership. The node retains the lease through the partition
+ * and must re-establish membership before it can release (or an operator can
+ * force-release via forceReleaseLeaseStep). */
 pred leaseReleaseStep[n: Node, s: Shard, t: Time, tnext: Time] {
     -- Preconditions
     holdsLease[n, s, t]
-    
+    isMember[n, t]
+    not hasFlapped[n, t]
```
udnay, line 1600: So if at time `t` the node has flapped it can't release its lease?

specs/coordination.als
```diff
@@ -2046,13 +2051,21 @@ assert splitChildrenContiguous {
...
+ * [SILO-COORD-INV-4] A partitioned node keeps serving but cannot mutate leases.
+ *
+ * A node that is partitioned from the coordination layer (flapped) continues to
+ * serve requests for shards it already owns — its permanent lease persists and
+ * holding the lease is sufficient authority to serve. It *cannot* acquire or
+ * release shard leases while partitioned: lease mutations require current
+ * membership, so ownership changes must wait until the node re-establishes
+ * membership (or an operator force-releases from a live node).
+ */
+assert flappedNodeCannotMutateLeases {
```
udnay, line 2063: 👍

udnay (approved):
This lines up with the k8s partition as well

specs/coordination.als
kirinrastogi: yes, it can't reach the api to delete the shard lease

PR #241, opened 2026-04-22
Resume interrupted shard splits on pod startup

kirinrastogi:
A pod that crashed mid-split while its split record was in SplitCloning used to stay stuck forever: on restart it logged "re-hydrated paused state" and did nothing else, and any subsequent split attempt on the same parent failed with "a split is already in progress". This was reproduced in a production gameday.

Instead, the pod should resume splits in progress and drive them to completion, otherwise the shard will be down. This requires a refactor to clone_closed_database so that the checkpoint id can be persisted in the k8s configmap.

Crash windows:
 - before prepare_split_checkpoint: checkpoint_id stays None; resumer re-enters cleanly, no leaks.
 - between prepare and store_split: checkpoint is written to the parent manifest but the UUID isn't persisted. Resumer creates a new one and proceeds. The first checkpoint is orphaned in the parent (harmless leak; eventual post-split GC would reclaim it).
 - after store_split: Some(id) persisted; resumer reuses it and slatedb's idempotent create_clone does the rest.

If there is a race, an operator will need to call siloctl to delete a checkpoint for a given shard.

Tests:
 - factory: clone_closed_shard_is_idempotent_with_same_checkpoint verifies the slatedb property the resumer relies on.
 - coordination: the existing "abandons split on crash" tests are rewritten as "resumes split on crash" (SplitPausing and SplitCloning). A k8s test verifies splits initiated by a different node are skipped.

### DIFF
--- a/specs/coordination.als
+++ b/specs/coordination.als
@@ -1505,24 +1505,15 @@ pred nodeFlapStep[n: Node, t: Time, tnext: Time] {
     -- Permanent leases: all leases are preserved through flap
     leaseFrame[t, tnext]
 
-    -- All incomplete splits are abandoned when the node crashes.
-    -- The shard map update is the commit point - if children don't exist in the shard map,
-    -- the split never completed and is safe to abandon. Orphaned clone databases may exist
-    -- but don't affect correctness.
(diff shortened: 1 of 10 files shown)


kirinrastogi:
closing because we can just siloctl to abandon a shard split if the pod dies

PR #243, opened 2026-04-23
Scaffold Standalone Compactor

kirinrastogi:
It is meant to run as a Deployment. They coordinate with each other and sort themselves based on pod names to acquire ranges of shards. They trample each other on scale up or down, but that is by design. Compactors can fence each other, and this is easier than maintaining a new lease type.

Tested locally in orbstack. The compactors run size tiered compaction with no filter currently.

It needs to be a new crate because compiling with the `compaction_filter` feature flag may affect snapshot isolation. It will soon have this feature flag once the compaction filter is implemented.

Still need to do load testing in staging or prod to see if the compactor can keep up with the writes based on our slatedb config.

Rollout plan is to add rbac and Deployment as documented in local-test dir, and to disable the embedded compactor by removing the settings for it: https://github.com/gadget-inc/silo/pull/242


silo-compactor/src/assignment.rs
```diff
@@ -0,0 +1,39 @@
...
+    shards.sort_by_key(|s| s.0);
+    shards.dedup();
+
+    let Some(my_idx) = pods.iter().position(|n| n == self_name) else {
+        return Vec::new();
+    };
+
+    let total_pods = pods.len();
+    let n = shards.len();
+    if n == 0 || total_pods == 0 {
```
udnay, line 28: /nit why not do this at the top of the function?

udnay (commented):
This looks ok to me. I wonder if we could just have the silo workers themselves trigger a compaction to a stable/specific compactor pod or something? Could potentially eliminate some of the shard-loader logic. It gets weird once compaction shards start to autoscale so maybe it doesn't make sense

silo-compactor/src/assignment.rs
kirinrastogi: will fix in a follow up

kirinrastogi:
we could do it that way, but I think it's easier this way because the compaction is modelled as a long running task that sleeps if there is nothing to do

PR #246, opened 2026-04-24
Add CompletedJobCompactionFilter to silo-compactor

kirinrastogi:
Enable slatedb's `compaction_filters` feature in silo-compactor and add a completed-jobs filter so the standalone compactor can purge terminal (Succeeded / Failed / Cancelled) job data past its retention window.

## What the filter drops

Six key prefixes are inspected during compaction:

| Prefix | Name | Decision method |
| --- | --- | --- |
| 0x01 | JOB_INFO | look up `(tenant, job_id)` in the pre-scanned set |
| 0x02 | JOB_STATUS | decode value, check `is_terminal && changed_at_ms < cutoff` |
| 0x03 | IDX_STATUS_TIME | self-contained: status + ts encoded in key |
| 0x04 | IDX_METADATA | look up `(tenant, job_id)` in the pre-scanned set |
| 0x07 | ATTEMPT | look up `(tenant, job_id)` in the pre-scanned set |
| 0x0A | JOB_CANCELLED | look up `(tenant, job_id)` in the pre-scanned set |

When `is_dest_last_run` is true the filter emits `Drop`; otherwise `Modify(Tombstone)`. Retention defaults to 7 days and is configurable via `[compaction_filter] kind = "completed_jobs" retention_secs = <n>`.

## Pre-scan optimization

Four of the six prefixes used to require a `JOB_STATUS` point read per row to determine the owning job's terminal status. With per-job ATTEMPT and per-key/value IDX_METADATA fan-out, that read amplification dominates compaction cost on busy shards.

Instead, at filter-creation time the supplier opens a short-lived `DbReader` and runs `build_expired_set` over the IDX_STATUS_TIME index (key prefix `0x03`). The index already encodes `(tenant, status, inverted_ts, job_id)` in its key, and silo's `set_job_status_with_index` in `src/job_store_shard/enqueue.rs:616` deletes the prior row on every status transition — so a single `DbReader::scan_prefix` returns one row per `(tenant, job_id)` reflecting current status. Verified via `restart.rs:149`, `import.rs:620`, and `delete_job` at `mod.rs:798`.

The result is an `ExpiredSet` of `(tenant, job_id)` pairs that the filter knows to be terminal + older than the cutoff. Per-row `is_job_expired` calls become O(1) HashMap lookups instead of one slatedb point read each.

## Memory bound

The set is stored as a per-tenant `HashMap<String, HashSet<String>>` so `contains(&str, &str)` doesn't allocate a tuple key on every lookup. It is bounded by a soft cap (`DEFAULT_EXPIRED_SET_MAX_ENTRIES = 250_000`, configurable via `expired_set_max_entries` in TOML or `with_max_expired_entries` in code; `0` disables the pre-scan entirely). If the IDX_STATUS_TIME pass would exceed the cap, `build_expired_set` returns `Ok(None)` and the filter falls back to per-row point reads — memory does not grow past the cap regardless of how many expired jobs exist in the index.

The set holds only expired jobs (not all jobs), and is dropped when the `CompactionFilter` is dropped after `on_compaction_end`. No process-lifetime accumulation. At ~100-150 bytes per entry the cap is ~37 MB per filter; worst-case 12 shards × 4 max_concurrent_compactions ≈ 1.8 GB across a pod, typically much less.

Owned `(String, String)` rather than 64-bit hashes — a hash collision would silently drop live data, which is not self-correcting; the cap bounds the absolute footprint anyway.

## Counter drift (follow-up)

Job counters (per-tenant `tenant_status_counter`, shard `total_jobs` / `completed_jobs`) are intentionally NOT touched by this filter. They will drift relative to the rows this filter drops, and they may stay stale for some time after compaction. A follow-up PR will introduce a background reconciliation task that scans live counters against live JOB_STATUS rows and corrects drift.

The standalone compactor process does not have a writable `Db` handle, so writing counter deltas inline from `on_compaction_end` would require opening a fenced writer alongside the live silo server — out of scope here.

## Block cache (future)

The pre-scan reads object-store blocks from the IDX_STATUS_TIME range once per compaction job. silo-compactor does not currently configure a block cache, so those reads do not benefit from warm-cache reuse across compaction jobs on the same shard. If profiling shows the pre-scan dominates compaction wall time, configuring a `with_db_cache` on the `DbReader` (and optionally on the Compactor) would be a natural next step. Out of scope for this PR.

## Other changes

- `silo-compactor` depends on the `silo` crate as a path dep with `default-features = false` to reuse `silo::keys`, `silo::codec`, `silo::job`, and `silo::job_store_shard::counter_merge_operator`.
- `silo::keys::prefix` is promoted from `pub(crate)` to `pub` so the filter can read the shared prefix bytes.
- `slatedb` and `slatedb-common` are pinned to 0.12.0; the `compaction_filters` feature is enabled on slatedb only in silo-compactor.
- Worker threads `expired_set_max_entries` through to the supplier via the new `with_max_expired_entries` builder.

## Test coverage

- 26 unit tests in `silo-compactor` (config + filter), including:
  - `ExpiredSet` semantics (`contains` by `&str`, dedup on insert).
  - `build_expired_set` against a real in-memory slatedb seeded with terminal+old, terminal+recent, non-terminal+old, non-terminal+recent rows across two tenants.
  - Cap-exceeded path returns `Ok(None)`; `max_entries = 0` short-circuits without scanning.
  - `is_job_expired` answers from the set when present without falling back to a point read.
  - Drop vs Tombstone decision based on `is_dest_last_run`.
- 1 end-to-end integration test (`silo-compactor/tests/compaction_filter_integration.rs`) that writes a mix of old-terminal, recent-terminal, and old-non-terminal rows to an in-memory SlateDB, runs a standalone Compactor with the filter, polls until L0 SSTs are consumed, and verifies that only expired-terminal rows were purged. Covers the Drop branch end-to-end. Tombstone semantics are covered at the unit level — the end-to-end tombstone scenario requires back-to-back L0 compaction submits via Admin, which slatedb 0.12.0 rejects as `InvalidCompaction` due to stale store snapshots between sequential admin submits.

### DIFF
--- a/silo-compactor/src/compaction_filter.rs
+++ b/silo-compactor/src/compaction_filter.rs
@@ -18,11 +18,11 @@
 //! | 0x07   | ATTEMPT          | Point-read JOB_STATUS via DbReader   |
 //! | 0x0A   | JOB_CANCELLED    | Point-read JOB_STATUS via DbReader   |
 //!
-//! In the standalone `silo-compactor` process we do not have a writable `Db`
-//! handle, so counter decrements are tracked but emitted only via the warn
-//! log on `on_compaction_end`. The live silo server owns the counters; those
-//! will drift slightly and are expected to be reconciled by silo's own
-//! cleanup paths.
+//! Job counters (per-tenant `tenant_status_counter`, shard
+//! `total_jobs`/`completed_jobs`) are NOT touched by this filter. They will
+//! drift relative to the keys this filter drops; reconciliation is
+//! delegated to a separate background task that scans the live counters
+//! against the live JOB_STATUS rows.
 
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
@@ -34,7 +34,7 @@ use silo::codec::decode_job_status_owned;
 use silo::job_store_shard::counter_merge_operator;
 use silo::keys::{
     idx_status_time_all_prefix, parse_attempt_key, parse_job_cancelled_key, parse_job_info_key,
-    parse_job_status_key, parse_metadata_index_key, parse_status_time_index_key, prefix,
+    parse_metadata_index_key, parse_status_time_index_key, prefix,
 };
 use slatedb::{
     CompactionFilter, CompactionFilterDecision, CompactionFilterError, CompactionFilterSupplier,
@@ -273,7 +273,6 @@ impl CompactionFilterSupplier for CompletedJobCompactionFilterSupplier {
             dropped_idx_metadata: 0,
             dropped_attempt: 0,
             dropped_job_cancelled: 0,
-            tenant_status_drops: HashMap::new(),
         }))
     }
 }
@@ -301,9 +300,6 @@ pub struct CompletedJobCompactionFilter {
     dropped_idx_metadata: u64,
     dropped_attempt: u64,
     dropped_job_cancelled: u64,
-
-    // (tenant, status_kind) → count for counter decrements.
-    tenant_status_drops: HashMap<(String, String), u64>,
 }
 
 impl CompletedJobCompactionFilter {
@@ -400,12 +396,6 @@ impl CompactionFilter for CompletedJobCompactionFilter {
                 };
 
                 if status.is_terminal() && status.changed_at_ms < self.cutoff_ms {
-                    if let Some(parsed) = parse_job_status_key(&entry.key) {
-                        *self
-                            .tenant_status_drops
-                            .entry((parsed.tenant, status.kind.as_str().to_owned()))
-                            .or_insert(0) += 1;
-                    }
                     self.dropped_status += 1;
                     return Ok(self.drop_decision());
                 }
@@ -476,23 +466,6 @@ impl CompactionFilter for CompletedJobCompactionFilter {
             );
         }
 
-        // In the standalone compactor we have no writable Db handle, so
-        // counter decrements are only tracked for diagnostics and emitted
-        // here. The live silo writer reconciles counters elsewhere.
-        if self.dropped_status > 0 {
-            let tenants: Vec<_> = self
-                .tenant_status_drops
-                .iter()
-                .map(|((tenant, status), count)| format!("{tenant}/{status}={count}"))
-                .collect();
-            warn!(
-                jobs_dropped = self.dropped_status,
-                tenant_status = ?tenants,
-                "completed_jobs compaction filter: counter decrements skipped \
-                 (standalone compactor has no Db writer)"
-            );
-        }
-
         Ok(())
     }
 }
@@ -541,7 +514,6 @@ mod tests {
             dropped_idx_metadata: 0,
             dropped_attempt: 0,
             dropped_job_cancelled: 0,
-            tenant_status_drops: HashMap::new(),
         }
     }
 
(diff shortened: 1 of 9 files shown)


silo-compactor/src/compaction_filter.rs
```diff
@@ -0,0 +1,760 @@
...
+            + self.dropped_status
+            + self.dropped_idx_status_time
+            + self.dropped_idx_metadata
+            + self.dropped_attempt
+            + self.dropped_job_cancelled
+    }
+
+    /// Returns [`Drop`] when the destination is the oldest sorted run (no
+    /// resurrection risk), [`Modify(Tombstone)`] otherwise.
+    fn drop_decision(&self) -> CompactionFilterDecision {
```
kirinrastogi, line 317: with size tiered compaction we see about 54% of the simulated drops become tombstones, with leveled compaction it is 0% because they are all dropping from the last sorted run.

however this is still better than a background task that deletes jobs because that would insert tombstones 100% of the time

silo-compactor/src/compaction_filter.rs
```diff
@@ -0,0 +1,760 @@
...
+/// Terminal status strings as stored in IDX_STATUS_TIME keys.
+const TERMINAL_STATUSES: &[&str] = &["Succeeded", "Failed", "Cancelled"];
+
+/// Default retention window for completed jobs: 7 days.
+pub const DEFAULT_RETENTION_SECS: u64 = 7 * 24 * 60 * 60;
+
+/// Default upper bound on the pre-scanned `ExpiredSet`. If the IDX_STATUS_TIME
+/// pass would exceed this, the supplier falls back to per-row point reads
+/// rather than holding an arbitrarily large set in memory.
+pub const DEFAULT_EXPIRED_SET_MAX_ENTRIES: usize = 250_000;
```
kirinrastogi, line 54: This is a guard to prevent OOMs in the compactor. The expired set should be a small subset of the data populated from the secondary index. Napkin math says that this will be at most 1-2GB per shard.

silo-compactor/src/compaction_filter.rs
```diff
@@ -0,0 +1,760 @@
...
+                            cap = self.expired_set_max_entries,
+                            "completed_jobs filter: pre-scan ready"
+                        );
+                        Some(Arc::new(set))
+                    }
+                    Ok(None) => {
+                        warn!(
+                            cap = self.expired_set_max_entries,
+                            elapsed_ms = started.elapsed().as_millis() as u64,
+                            "completed_jobs filter: pre-scan exceeded cap or disabled, \
```
kirinrastogi, line 247: I will add an alert on this, it means that this is falling back to point reads instead of using the pre-assembled set

silo-compactor/src/compaction_filter.rs
```diff
@@ -0,0 +1,760 @@
...
+                }
+            }
+
+            _ => {}
+        }
+
+        Ok(CompactionFilterDecision::Keep)
+    }
+
+    async fn on_compaction_end(&mut self) -> Result<(), CompactionFilterError> {
```
kirinrastogi, line 451: originally I wanted to update counters here, but it is not possible with standalone compaction, we would need to write to the db. So instead it will be a background task in silo that fixes the counter staleness - to be done in a follow up PR.

PR #233, opened 2026-04-12
Optimize grant scanner: batch writes and concurrent status validation

airhorns:
## Summary

- Restructures `process_grants` from per-grant sequential I/O to a batched pipeline that does a single durable write for all grants
- Batch-validates job status keys concurrently via `futures::join_all` instead of sequential `db.get` per request
- Stale request deletions are folded into the same write batch instead of separate writes
- Adds `process_grants` benchmarks at 5/50/500 grant scales to measure the grant scanner hot path

## Benchmark Results

| Benchmark | Before (per-grant writes) | After (batched) | Speedup |
|---|---|---|---|
| `process_grants(5)` | 11.6ms | 2.8ms | **4.1x** |
| `process_grants(50)` | 118.5ms | 1.6ms | **74x** |
| `process_grants(500)` | 1,140ms | 7.9ms | **144x** |

The improvement scales super-linearly: the old code does N separate durable writes (each waiting for WAL flush), the new code does exactly 1 regardless of batch size.

## How it works

The new `process_grants` pipeline:
1. **Scan & decode** — iterate request queue, decode candidates, resolve concurrency limit once
2. **Batch validate** — issue all job status reads concurrently, partition into valid vs stale
3. **Reserve & build batch** — try in-memory slot reservations for valid candidates, accumulate all edits into a single `WriteBatch`
4. **Single durable write** — one `write_with_options(await_durable: true)` for all grants + stale deletions

## Test plan

- [x] All 298 core shard tests pass (enqueue, dequeue, import, cancel, lease, concurrency, retry, query, cleanup)
- [x] `reimport_failed_to_scheduled_ignores_stale_concurrency_requests` passes (validates stale cleanup still works when queue is at capacity)
- [x] New benchmarks run successfully at all three scales
- [ ] CI passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Touches core concurrency grant logic and changes request cleanup/counter updates, so bugs could impact capacity enforcement or task scheduling despite added benchmarks/tests.
> 
> **Overview**
> **Grant scanning throughput is significantly improved** by rewriting `ConcurrencyManager::process_grants` into a batched pipeline: scan candidate requests in batches, validate job statuses concurrently, reserve in-memory slots, and commit all holder/task creations plus stale/corrupt request deletions in one durable `WriteBatch` (with a single combined requester-counter decrement).
> 
[...]

PR #252, opened 2026-04-30
Bound grant scanner status-lookup fan-out and per-call batch size

kirinrastogi:
This join_all was introduced in https://github.com/gadget-inc/silo/pull/233 but when there are many outstanding grants to process this causes OOMs, so we need to chunk it.

Heap profiles from production showed `process_grants` pinning multi-GB of `Bytes` in `CachedObjectStore::read_part` via an unbounded `futures::future::join_all` over `db.get(job_status_key)` lookups. With large pending-request backlogs, each `db.get` fans out to many slatedb SST block fetches, and thousands in flight at once exceeds container memory before a single pass can complete (19 OOMKills on silo-1, similar on silo-5/6).

- Replace `join_all` with `stream::iter(...).buffered(64)` — order-preserving so the `scanned.zip(status_results)` pairing still holds, while capping concurrent slatedb reads.
-  ~In `start_grant_scanner`, chunk each queue's `count` into batches of at most 1024 per `process_grants` call. Bounds WriteBatch growth and releases memory between durable writes; short-circuits when a chunk grants fewer than asked (iterator/capacity exhausted).~ removed because it was causing GRPC timeouts / client cancellations
- Add a regression test that drains a 200-waiter backlog through the buffered pipeline across multiple in-flight windows.

---

process_grants benchmark results

process_grants benchmark — branch vs main

```  ┌───────────────────────┬───────────────────────────┬───────────────────────┬────────────────┐
  │ Case (grants/waiters) │ main (join_all unbounded) │ branch (buffered(64)) │       Δ        │
  ├───────────────────────┼───────────────────────────┼───────────────────────┼────────────────┤
  │ small (5 / 20)        │ 1.5 ms                    │ 2.2 ms                │ +0.7 ms (+47%) │
  ├───────────────────────┼───────────────────────────┼───────────────────────┼────────────────┤
  │ medium (50 / 200)     │ 1.9 ms                    │ 1.8 ms                │ −0.1 ms (−5%)  │
  ├───────────────────────┼───────────────────────────┼───────────────────────┼────────────────┤
  │ large (500 / 1000)    │ 8.9 ms                    │ 6.6 ms                │ −2.3 ms (−26%) │
  └───────────────────────┴───────────────────────────┴───────────────────────┴────────────────┘
```

  Interpretation

  - Large case (500 grants): branch is ~26% faster. This is the case where bounded concurrency could matter: with join_all unbounded, 500 simultaneous db.get(job_status_key) calls fan out to many in-flight slatedb
  block fetches, increasing cache/memory pressure. Capping at 64 in-flight reads appears to actually help throughput here, in addition to bounding heap residency.
  - Medium case (50): essentially identical — 50 < 64 cap, so buffered and join_all schedule the same number of futures.
  - Small case (5): branch shows +0.7 ms, but this is a single-iteration measurement (bench_op is called with 0 warmup + 1 measured for process_grants), so a sub-ms delta on a 1.5 ms run is well within run-to-run
  noise. Each invocation also uses a freshly cloned shard, so there's no shared cache state between branches.

---

another benchmark with process_grants called 10x for the 500 case

 process_grants(500) — 10 iterations each

```
  Per-iteration timings (ms)

  ┌──────┬─────────────────┬───────────────────────┐
  │ iter │ main (join_all) │ branch (buffered(64)) │
  ├──────┼─────────────────┼───────────────────────┤
  │    0 │            8.40 │                  6.09 │
  ├──────┼─────────────────┼───────────────────────┤
  │    1 │           11.59 │                  6.81 │
  ├──────┼─────────────────┼───────────────────────┤
  │    2 │            7.29 │                  7.50 │
  ├──────┼─────────────────┼───────────────────────┤
  │    3 │            6.23 │                  5.44 │
  ├──────┼─────────────────┼───────────────────────┤
  │    4 │           10.27 │                  6.67 │
  ├──────┼─────────────────┼───────────────────────┤
  │    5 │            6.66 │                  6.30 │
  ├──────┼─────────────────┼───────────────────────┤
  │    6 │            7.45 │                  6.11 │
  ├──────┼─────────────────┼───────────────────────┤
  │    7 │            6.59 │                  6.48 │
  ├──────┼─────────────────┼───────────────────────┤
  │    8 │            9.81 │                  6.20 │
  ├──────┼─────────────────┼───────────────────────┤
  │    9 │           10.37 │                  8.54 │
  └──────┴─────────────────┴───────────────────────┘

  Aggregated

  ┌────────┬─────────┬────────┬──────┐
  │ metric │  main   │ branch │  Δ   │
  ├────────┼─────────┼────────┼──────┤
  │ min    │  6.2 ms │ 5.4 ms │ −13% │
  ├────────┼─────────┼────────┼──────┤
  │ p50    │  8.4 ms │ 6.5 ms │ −23% │
  ├────────┼─────────┼────────┼──────┤
  │ mean   │  8.5 ms │ 6.6 ms │ −22% │
  ├────────┼─────────┼────────┼──────┤
  │ p99    │ 11.6 ms │ 8.5 ms │ −27% │
  ├────────┼─────────┼────────┼──────┤
  │ max    │ 11.6 ms │ 8.5 ms │ −27% │
  └────────┴─────────┴────────┴──────┘

```
 
 Throughput (grants/sec, mean)

  - main: 500 / 0.0085 ≈ 58.8 K grants/sec
  - branch: 500 / 0.0066 ≈ 75.8 K grants/sec
  - +29% on the bounded-concurrency branch

### DIFF
--- a/src/concurrency.rs
+++ b/src/concurrency.rs
@@ -40,6 +40,8 @@ use std::collections::{BTreeMap, HashMap, HashSet};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 
+use futures::StreamExt;
+
 use slatedb::config::WriteOptions;
 use slatedb::{Db, DbIterator, WriteBatch};
 
@@ -83,6 +85,13 @@ impl From<crate::codec::CodecError> for ConcurrencyError {
         ConcurrencyError::Encoding(e.to_string())
     }
 }
+
+/// Max in-flight `db.get(job_status_key)` lookups during a single
+/// `process_grants` pass. Caps slatedb-side block-fetch fan-out so the
+/// scanner can't pin unbounded `Bytes` while validating a large pending
+/// backlog. Mirrors the `CONCURRENCY = 64` pattern used in
+/// `job_store_shard::get_jobs_status_batch` for the same reason.
+const STATUS_LOOKUP_CONCURRENCY: usize = 64;
 
 /// Result of attempting to enqueue a job with concurrency limits
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


src/concurrency.rs
```diff
@@ -917,14 +948,25 @@ impl ConcurrencyManager {
...
-                })
+                .map(|c| job_status_key(tenant, &c.job_id))
                 .collect();
-            let status_results = futures::future::join_all(status_futures).await;
+            let status_results: Vec<_> = futures::stream::iter(
+                status_keys
+                    .into_iter()
+                    .map(|key| async move { db.get(&key).await }),
+            )
+            .buffered(STATUS_LOOKUP_CONCURRENCY)
```
kirinrastogi, line 945: I think this can be conceptualized as a pMap of 64

udnay (approved):

PR #255, opened 2026-05-02
When workers crash and their leases expire we leak holders

udnay:
Problem

  In production, floatingConcurrency holder rows accumulate to exactly defaultMaxConcurrency per (tenant, queue), with job_id = NULL (the corresponding job is gone). Once the cap is hit, no new requesters are admitted — the limit is permanently
  saturated. One observed tenant: 540 stranded holders spanning a ~7.5h window, with 588k requesters piled up waiting.

  Root cause (one path proven)

  JobStoreShard::handle_run_attempt in src/job_store_shard/dequeue.rs looks up job_info for the task it's about to lease. If job_info is missing it deletes the task and returns Ok — without releasing the concurrency holders the task acquired at
  enqueue time. The holder was created atomically with the task at enqueue, but the cleanup only handled the task half:

  let Some(job_bytes) = maybe_job else {
      state.batch.delete(task_key);  // task: gone
      state.ack_deleted(task_key);
      return Ok(());                  // holder: still there, forever
  };

  Once we land in this branch, the holder is unreachable to every other release path: the reaper only walks lease records (no lease was ever created), report_attempt_outcome only releases holders attached to a lease (none), and the grant scanner
   only creates holders. Repeat per affected job until the cap is saturated.

  I have not nailed down what causes job_info to go missing on the owning shard in production. The leak path itself is reproduced deterministically in a unit test by manually deleting job_info_key after enqueue and before dequeue. The fix is
  defensible regardless of the upstream cause, since the dequeue path shouldn't strand a holder under any precondition failure.

  Fix

  src/job_store_shard/dequeue.rs — handle_run_attempt's missing-job branch now deletes each concurrency_holder_key(tenant, queue, task_id) for every entry in the RunAttempt's held_queues in the same batch as the task delete. A new
  holder_releases accumulator on DequeueIterationState carries these to a post-commit step that runs atomic_release + request_grant for each, dropping the in-memory cap and waking the grant scanner so a queued requester takes the freed slot
  immediately. Includes a tracing::warn! so this path is observable if it fires.

  src/job_store_shard/lease.rs — added JobStoreShard::purge_orphaned_holders_for_task(tenant, task_id). Tenant-scoped scan (bounded by the cap) that deletes any holder rows attributed to task_id and post-commit calls atomic_release +
  request_grant. Idempotent.

  src/server.rs — when the gRPC report_outcome returns LeaseNotFound, the handler now invokes purge_orphaned_holders_for_task (when tenant_id is on the request) before returning the error. This is the "defensive idempotent cleanup" hook so that
  a worker hitting TaskNotFoundError on a retry causes any matching stranded holder to be cleaned up. Belt-and-suspenders for any leak path I haven't found.

  Tests (tests/holder_leak_tests.rs, all passing)

  - holder_leaked_when_run_attempt_finds_missing_job_info — the direct reproduction. Fails before the fix, passes after.
  - floating_concurrency_holder_released_on_lease_expiry — invariant: the simple reaper path releases the holder.
  - floating_concurrency_holder_released_when_report_outcome_after_reap — invariant: a late ack after reap returns LeaseNotFound and leaves no holder.
  - purge_orphaned_holders_for_task_removes_stranded_holders — the purge primitive: removes target holders, leaves unrelated ones, idempotent.
  - late_report_outcome_followed_by_purge_clears_holder — end-to-end: late ack + purge clears a planted orphan.
  - floating_concurrency_steady_state_zero_holders_after_full_cycle — drives multiple jobs through enqueue → lease → mixed (success, reap) and asserts zero holders remain.

  Each test ends with count_concurrency_holders == 0 as a steady-state invariant so future regressions in any release path will fire here.

  Out of scope (flagged for follow-up)

  I found two pre-existing bugs while investigating. Neither matches the production fingerprint (gadget uses a single FloatingConcurrency limit, not a chain), so I did not fix them in this PR:

  - Chained-limit advancement is broken at src/job_store_shard/enqueue.rs:67-80. record_grant_outcome swaps the None/Some arms relative to what concurrency::handle_enqueue returns, so any job with multiple limits stops at the first grant. Easy
  reproduction: [Concurrency, Concurrency] produces 1 holder instead of 2.
  - CheckRateLimit max-retries silently drops held_queues at src/job_store_shard/dequeue.rs:534-541. Dormant today (chains never advance, so CheckRateLimit tasks never accumulate held_queues), but will leak as soon as the chain bug is fixed.

  These should be fixed together in a follow-up.

  Verification

  - cargo test --test holder_leak_tests — 6/6 pass.
  - cargo test --test concurrency_tests --test floating_concurrency_tests — 22/22 and 21/21 still pass.
  - cargo test --tests — all pass except cluster_query_integration_tests which require a local etcd; unrelated to these changes.

  After merge, query production for entry_type = 'holder' AND job_id IS NULL. The leak rate on new traffic should be zero. Existing leaked holders for affected tenants need a one-time cleanup (not done here — out of scope).

  Risk

  - The new dequeue post-commit step adds an atomic_release + request_grant call per holder released in the missing-job branch. Hot path is unchanged for the normal case (the loop is empty when no holders were dropped).
  - purge_orphaned_holders_for_task is only called from the LeaseNotFound error path on report_outcome. Tenant-scoped scan, bounded by the cap (hundreds of holders typical), so cost is acceptable for an exception path.
  - No protocol changes. No client changes required.

---

> [!NOTE]
> **Medium Risk**
> Touches concurrency holder lifecycle across dequeue and outcome reporting; mistakes could cause incorrect capacity accounting or extra DB work on error paths, but changes are mostly defensive and isolated to cleanup branches.
> 
> **Overview**
> Fixes a holder-leak where `RunAttempt` tasks dropped during dequeue (missing `job_info`) could strand `FloatingConcurrency` holder rows and permanently consume queue capacity.
> 
> `dequeue` now deletes any `concurrency_holder_key`s associated with a dropped task and performs a post-commit in-memory `atomic_release` + `request_grant` wakeup to immediately free capacity. Adds a new `purge_orphaned_holders_for_task` API that tenant-scans and deletes holder rows for a `task_id`, and the gRPC `report_outcome` handler invokes it on `LeaseNotFound` as best-effort self-healing.
> 
> Adds `tests/holder_leak_tests.rs` covering lease-reap, missing-job-info drop, late outcome reporting, purge idempotency, and steady-state “zero holders” invariants.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c5b6381e71c6fcee820710feb661ac9412fcda42. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/job_store_shard/dequeue.rs
+++ b/src/job_store_shard/dequeue.rs
@@ -11,7 +11,9 @@ use crate::job::{JobStatus, JobView};
 use crate::job_attempt::{AttemptStatus, JobAttempt, JobAttemptView};
 use crate::job_store_shard::helpers::{DbWriteBatcher, now_epoch_ms};
 use crate::job_store_shard::{DequeueResult, JobStoreShard, JobStoreShardError, LimitTaskParams};
-use crate::keys::{attempt_key, job_info_key, leased_task_key, parse_task_key};
+use crate::keys::{
+    attempt_key, concurrency_holder_key, job_info_key, leased_task_key, parse_task_key,
+};
 use crate::shard_range::ShardRange;
 use crate::task::{DEFAULT_LEASE_MS, LeaseRecord, LeasedRefreshTask, LeasedTask, Task};
 use crate::task_broker::BrokerTask;
@@ -26,6 +28,10 @@ struct DequeueIterationState {
     grants_to_rollback: Vec<(String, String, String)>,
     leased_tasks_for_dst: Vec<(String, String, String)>,
     pending_attempts: Vec<(String, JobView, Vec<u8>)>,
+    /// Holders that were deleted in this iteration's batch and need their
+    /// in-memory slot released + the grant scanner notified after commit.
+    /// Format: (tenant, queue, task_id).
+    holder_releases: Vec<(String, String, String)>,
     processed_internal: bool,
 }
 
@@ -38,6 +44,7 @@ impl DequeueIterationState {
             grants_to_rollback: Vec::new(),
             leased_tasks_for_dst: Vec::new(),
             pending_attempts: Vec::new(),
+            holder_releases: Vec::new(),
             processed_internal: false,
         }
     }
@@ -84,6 +91,9 @@ impl JobStoreShard {
         // Track leased tasks for DST event emission after commit
         // Format: (tenant, job_id, task_id)
         let mut leased_tasks_for_dst: Vec<(String, String, String)> = Vec::new();
+        // Track holders deleted in the batch that need post-commit in-memory
+        // release + grant-scanner wakeup. Format: (tenant, queue, task_id).
+        let mut holder_releases: Vec<(String, String, String)> = Vec::new();
 
         // Loop to process internal tasks until we have tasks that are destined for the worker, or no more ready tasks at all.
         const MAX_INTERNAL_ITERATIONS: usize = 10;
@@ -173,6 +183,7 @@ impl JobStoreShard {
 
             // Merge iteration state into outer accumulators
             grants_to_rollback.append(&mut state.grants_to_rollback);
+            holder_releases.append(&mut state.holder_releases);
 
             // Two-phase DST events: emit before write for correct causal ordering,
             // confirm after write succeeds, cancel if write fails.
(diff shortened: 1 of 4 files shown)


udnay:

- #255 ◀

kirinrastogi (approved):

kirinrastogi:
force push was a rebase for a new container build without the memory leak

PR #269, opened 2026-05-04
DST coverage for the holder-leak bug class + Bug 2/3 fixes

kirinrastogi:
I strongly recommend reviewing with an agent's `/review` feature, the alloy spec is dense.

## Summary

- **Bug 2** (`ac305df`): `handle_check_rate_limit`'s max-retries early return dropped the task without releasing held concurrency queues. Symmetric to PR #255's `handle_run_attempt` fix.
- **Bug 3** (`ae70b12`): `handle_check_rate_limit`'s missing/malformed `job_info` early returns had the same shape. Found while writing the DST coverage and fixed in scope.
- **Spec extension** (`565e15a`, `257d39a`): `specs/job_shard.als` now models `CheckRateLimit` tasks and drop-without-completion transitions. `assert holdersRequireActiveTask` is now load-bearing — it actually catches a forgotten release. Verified with `alloy6 -c holdersRequireActiveTask` (UNSAT) and a fitness test that omits the release (SAT).
- **silo-sim** (`76e14da`, `257d39a`): mixes RateLimit / chained Concurrency+RateLimit / FloatingConcurrency in the workload, injects worker crashes and `job_info` deletions, runs a real lease reaper, drains 2s before the strict end-of-run check. Critical post-review fix: `count_with_prefix` was scanning ASCII `"holders/"` while real keys use binary `0x09`, making the mid-run and end-of-run invariants vacuous. Now uses `keys::*_prefix()` everywhere.

Background: prior to this PR the Alloy spec didn't model `CheckRateLimit` at all and the runtime simulator only enqueued `Limit::Concurrency`, never simulated worker crashes, never deleted `job_info`, and didn't run the reaper. None of the holder-leak bugs in production could be observed by either tool.

## Out of scope (deliberate)

- `record_grant_outcome` at `enqueue.rs:67-80` interprets `None`/`Some` backwards relative to what `concurrency::handle_enqueue` actually returns, so chained limits silently drop the second limit — this means a chained `[Concurrency, RateLimit]` job today never produces a CheckRateLimit task with non-empty `held_queues`, which is why Bugs 2 & 3 are dormant in production. Flagged in `tests/holder_leak_tests.rs`'s NOTE; the fixes here are defensive for when that's fixed. Separate PR.

## Test plan

- [x] `cargo test -p silo --test holder_leak_tests` — all 8 tests pass.
- [x] Bug 2 regression test fails without the fix (verified by reverting the release block and re-running — `left: 1, right: 0`).
- [x] Bug 3 regression test fails without the fix (same verification).
- [x] `alloy6 exec -c holdersRequireActiveTask specs/job_shard.als` returns UNSAT.
- [x] `alloy6 exec -c exampleCheckRateLimitDropReleasesHolder specs/job_shard.als` returns SAT (witness reachable).
- [x] `alloy6 exec -c exampleDequeueDropRunAttemptReleasesHolder specs/job_shard.als` returns SAT.
- [x] Fitness test: replacing `releaseHolders[...]` with `holdersUnchanged` in `dequeueDropCheckRateLimit` makes `holdersRequireActiveTask` go SAT — the spec catches the bug.
- [x] `cargo check --bin silo-sim` clean.
- [ ] End-to-end simulator run: `cargo run --release --bin silo-sim -- --duration_secs 90 --shards 4 --rate_limit_fraction 0.20 --chained_fraction 0.25 --floating_fraction 0.10 --worker_crash_prob 0.01 --job_info_delete_prob 0.005 --seed 42`. Requires etcd; please run in your environment to confirm. With all three fixes in place this should pass cleanly; reverting any of (Bug 1 PR #255, Bug 2, Bug 3) on a scratch branch should make it fail at the end-of-run holder-count assertion.
- [ ] Full Alloy verification: `node scripts/run-alloy-verification.mjs specs/job_shard.als`. The full suite runs every check and witness in the file at expanded scopes — takes a long time at the new scopes; run before merge.

## Notes for reviewers

- Commit `ac305df` is also the head of the parent branch `kirin/fix-rate-limit-holder-leak`. If you'd prefer to land that one first, this PR's base can be retargeted.
- The simulator changes are non-deterministic; the seed is echoed at startup so failures are reproducible.
- The spec's well-formedness change (moving the holder/active-task invariant from `fact` to `assert`) is structural — it means the assertion is now meaningful, but it also means the spec admits states that violate it. The Alloy `traces` fact ensures only states reachable by `step` are considered, so the assertion is checked exactly against the modeled transitions.

### DIFF
--- a/specs/job_shard.als
+++ b/specs/job_shard.als
@@ -251,20 +251,13 @@ fact wellFormed {
     -- Requests are only for queues the job requires
     all r: TicketRequest | r.tr_queue in jobQueues[r.tr_job]
     
-    -- A holder can only exist for a task that is either:
-    -- 1. In the DB queue as a RunAttempt (granted at enqueue), OR
-    -- 2. In the buffer as a RunAttempt (scanned from DB), OR
-    -- 3. Has a lease (running), OR
-    -- 4. In the DB queue as a CheckRateLimit task (chained Concurrency -> RateLimit), OR
-    -- 5. In the buffer as a CheckRateLimit task
-    -- This allows holders to exist before lease (granted at enqueue, leased at dequeue)
-    -- and persists across the CheckRateLimit chain phase.
-    all h: TicketHolder |
-        some dbQueuedAt[h.th_task, h.th_time] or
-        some bufferedAt[h.th_task, h.th_time] or
-        some leaseAt[h.th_task, h.th_time] or
-        some dbCheckRateLimitAt[h.th_task, h.th_time] or
-        some bufferedCheckRateLimitAt[h.th_task, h.th_time]
+    -- The holder/active-task invariant lives in `assert holdersRequireActiveTask`
+    -- (below), NOT here. Keeping it as a fact makes the assertion trivially
+    -- true and prevents Alloy from generating buggy traces — the assertion
+    -- could never catch a transition that drops a task without releasing
+    -- holders. Encoded as an assertion, the safety property is checked
+    -- against the modeled transitions: a forgotten release produces a
+    -- counterexample.
     
     -- A holder and request can never coexist for the same (task, queue) at the same time
     -- (This is enforced by transitions but stated explicitly for clarity)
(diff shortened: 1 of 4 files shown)


udnay (approved):
I reviewed this with claude and asked some questions. LGTM

PR #283, opened 2026-05-08
Periodically force-reopen stuck shard leases

kirinrastogi:
## Summary

When a pod's `factory.close()` hits its 30s timeout (e.g., hyperdisk-ha-backed WAL can't flush because object storage is unreachable), the shard lease is intentionally retained to prevent data loss. Previously the retry loop re-called `shard.close()` on the same half-closed SlateDB instance with no backoff — never made progress and could even silently release the lease via the `slatedb::ErrorKind::Closed` "treat as success" branch.

- New `factory.force_reopen(shard_id, range)` drops the cached entry, briefly drains the old shard, and re-opens fresh (SlateDB WAL recovery).
- Releasing branch in both etcd and k8s backends now applies exponential backoff (5s → 5min cap, deterministic per-shard jitter) and calls `force_reopen` before each close retry. Tracked as `release_attempts` + `next_release_attempt` on `ShardGuardState`; reset when `desired` flips back to true or release succeeds.
- New `ShardForceReopened` DST event; `ShardOwnershipTracker` sharpens close-before-release so a re-open invalidates the prior pending close.
- Alloy spec gains a `ShardCloseable` flag — `leaseReleaseStep` requires it, `closeAttemptFailedStep` clears it, `forceReopenStep` restores it. New asserts `closeableBeforeRelease` and `forceReopenPreservesLease` both verify UNSAT alongside the 24 existing checks.

## Test plan

- [x] `cargo check --features k8s,dst --all-targets` clean
- [x] `cargo fmt`
- [x] `factory_tests` (13/13, including 2 new `force_reopen` tests)
- [x] `coordination_mod_tests` (18/18, including 2 new backoff tests)
- [x] `cleanup_reacquire_tests` (10/10)
- [x] Library tests (4/4)
- [x] `alloy6 exec specs/coordination.als` — all 26 checks pass (9 example traces SAT, 24 existing assertions UNSAT, 2 new assertions UNSAT)
- [ ] `etcd_coordination_tests` — needs running etcd; verify locally with `process-compose up etcd`
- [ ] Manual smoke in process-compose: induce object-store unreachability, trigger ownership change, observe `force_reopen` retries with backoff in logs, restore object store, observe clean release

### DIFF
--- a/specs/coordination.als
+++ b/specs/coordination.als
@@ -126,6 +126,25 @@ sig ShardLease {
 }
 
 /**
+ * Marker that a shard's local SlateDB instance has finished flushing
+ * its WAL to durable storage and the lease holder can safely release.
+ *
+ * In the implementation this corresponds to a successful `factory.close`
+ * that completed within the close timeout. If close fails (e.g., object
+ * storage unreachable), the closeable marker is dropped — the lease
+ * holder must retry by force-reopening the shard (fresh SlateDB instance
+ * with WAL recovery) and then attempting close again.
+ *
+ * [SILO-COORD-INV-15] A node may only release a shard lease while the
+ * shard is closeable. This makes the close-before-release invariant
+ * explicit at the spec level.
+ */
+sig ShardCloseable {
+    closeable_shard: one Shard,
+    closeable_time: one Time
+}
+
+/**
  * Tracks an in-progress shard split operation.
  * The split creates two child shards from one parent.
  */
@@ -451,6 +470,14 @@ fun shardsHeldBy[n: Node, t: Time]: set Shard {
 /** Check if a shard has any lease holder at time t */
 pred hasLeaseHolder[s: Shard, t: Time] {
     some leaseHolder[s, t]
+}
+
+/**
+ * Check if a shard's data is closeable (flushed to durable storage and
+ * ready to release) at time t. See ShardCloseable.
+ */
+pred shardCloseable[s: Shard, t: Time] {
+    some c: ShardCloseable | c.closeable_shard = s and c.closeable_time = t
 }
 
 /**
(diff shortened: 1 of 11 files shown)


specs/coordination.als
```diff
@@ -1592,37 +1653,111 @@ pred leaseAcquireStep[n: Node, s: Shard, t: Time, tnext: Time] {
...
+ * spec and rules out releasing a lease while WAL data remains unflushed. */
 pred leaseReleaseStep[n: Node, s: Shard, t: Time, tnext: Time] {
     -- Preconditions
     holdsLease[n, s, t]
+    shardCloseable[s, t]
     isMember[n, t]
     not hasFlapped[n, t]
 
     -- Postconditions
     not hasLeaseHolder[s, tnext]
```
udnay, line 1670: Not from this change but I don't think this is strictly true, we can say that `not holdsLease[n, s, tnext]` but I'm not sure if we can say that at `tnext` all nodes won't hold a lease.

specs/coordination.als
```diff
@@ -1592,37 +1653,111 @@ pred leaseAcquireStep[n: Node, s: Shard, t: Time, tnext: Time] {
...
+ * Operator escape hatch for permanently lost nodes. Force-release does NOT
+ * require closeable (that's the whole point — the operator is overriding
+ * a stuck lease) and clears closeable on the way out. */
 pred forceReleaseLeaseStep[s: Shard, t: Time, tnext: Time] {
     -- Preconditions
     hasLeaseHolder[s, t]
 
     -- Postconditions
     not hasLeaseHolder[s, tnext]
+    not shardCloseable[s, tnext]
```
udnay, line 1692: Shouldn't this also be a precondition? I guess it's unknown if the shard is closeable or not if the node is dead. Ah, is that why it's not in the preconditions?

kirinrastogi:
going to try a cleaner approach

PR #293, opened 2026-05-10
Add row TTL on terminal job records (gated)

kirinrastogi:
## Summary

When `terminal_job_expire_ms` is set, jobs reaching a terminal status (Succeeded/Failed/Cancelled) have their associated KV records re-put with a SlateDB row TTL so they age out of the store via compaction. Default off (`None`) — opt-in per shard.

Records given a TTL on the terminal write path:
- `JOB_INFO` (re-put with TTL)
- `JOB_STATUS` and `IDX_STATUS_TIME` (the new terminal status, written via `set_job_status_with_index_opts`)
- `IDX_METADATA` (one entry per metadata pair on the job)
- `ATTEMPT` (prefix-scanned and re-put with TTL; the in-batch new attempt also gets the TTL)
- `JOB_CANCELLED` (re-put with TTL if present)

Counters and shard-global rows are intentionally not given a TTL.

A new bench, \`benches/job_termination.rs\`, runs the termination path with TTL on/off across consumer counts and metadata sizes for A/B comparison.

## Known semantic implications (not addressed in this PR)

The spec invariant at \`specs/job_shard.als:215\` ("once a job exists, it always exists") is violated when this feature is enabled: after \`expire_ts\` the JOB_INFO row is gone and \`restartFailedJob\` / \`reimport\` / \`delete_job\` will return \`JobNotFound\`. A follow-up will model terminal-job expiration in the spec and weaken the existence invariants accordingly.

## Test plan

- [x] \`cargo check --all-targets\` clean
- [x] \`cargo test --lib\`, lease-task and counter-reconcile integration tests pass
- [x] \`cargo bench --bench job_termination\` runs end-to-end on FS backend (overhead within noise on local FS, as expected — WAL fsync dominates)
- [ ] Run on staging with \`terminal_job_expire_ms = 60_000\` to confirm rows actually drop after the TTL elapses
- [ ] Re-run bench against object-storage-backed config to validate the production storage shape

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -162,6 +162,10 @@ required-features = ["server"]
 name = "task_scanner_profile"
 harness = false
 
+[[bench]]
+name = "job_termination"
+harness = false
+
 [[test]]
 name = "turmoil_runner"
 path = "tests/turmoil_runner/main.rs"
--- a/benches/bench_helpers.rs
+++ b/benches/bench_helpers.rs
@@ -612,7 +612,7 @@ pub async fn clone_golden_shard(
             concurrency_reconcile_interval: Duration::from_millis(
                 silo::settings::DEFAULT_CONCURRENCY_RECONCILE_INTERVAL_MS,
             ),
-            enable_counter_reconciliation: false,
+            counter_reconciliation_seconds: None,
             hydrate_all_at_startup: false,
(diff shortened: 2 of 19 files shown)


kirinrastogi:
shipped in other prs

PR #318, opened 2026-05-17
Plumb held_queues + limit_index through queued concurrency requests

kirinrastogi:
## Summary

PR #317's Concurrency-before-FloatingConcurrency reorder mitigated the common Gadget shape, but left a latent gap on any limit chain where an earlier limit grants and a later concurrency limit queues. On the queued-limit promotion path:

- The granted RunAttempt was built with `held_queues = [just-granted queue]` only, orphaning any earlier-accumulated holders → permanent leak on outcome release.
- The chain was never resumed, so any limit after the queued one was bypassed → RunAttempt became leasable while a downstream concurrency/rate limit was still full.

Both symptoms applied to `[C1, C2]` (C1 grants, C2 queues), `[C, FC]`, `[FC, FC]`, and any 3+ chain `[C1, C2, C3]` where a middle limit queued.

This change:

- Threads accumulated `held_queues` AND the canonical-order `limit_index` (+ `total_limits` for terminal-vs-resume decisions) through the queued `EnqueueTask` / `RequestTicket` encoding.
- Replaces `Vec<String>` held_queues with `Vec<HeldQueue { queue, task_id }>` so each holder is released under the task_id that *created* it, removing the implicit invariant that every holder in a chain shares one task_id (broken by the resume-after-grant path where the request_id differs from the final RunAttempt's id).
- Adds `Task::ResumeAfterConcurrencyGrant`. `process_grants` writes it when `limit_index + 1 < total_limits`; `handle_resume_after_grant` in dequeue drives the chain forward via the existing `enqueue_limit_task_at_index`.
- Updates `handle_request_ticket`'s grant path to also walk the chain, with `inline_terminal: true` so a terminal RunAttempt can be leased without leaving a stale DB task at task_key.

Release sites in `lease.rs`, `cancel.rs`, `dequeue.rs`, and `import.rs` now look up each holder via `concurrency_holder_key(tenant, &hq.queue, &hq.task_id)` so mixed-task_id chains release correctly. The `Vec<HeldQueue>` signature forces this change at every call site.

## Test plan

- [x] `cargo check --workspace --tests` clean
- [x] `cargo clippy --tests --no-deps` no new warnings
- [x] `cargo fmt` clean
- [x] New regression tests in `tests/job_store_shard_concurrency_tests.rs`:
  - `chain_c1_c2_promotion_preserves_prior_holder_and_releases_both` — `[C1, C2]` grant-then-queue: lease has both holders, both release on success.
  - `chain_3_limits_middle_queue_does_not_bypass_later_limits` — `[C1, C2, C3]` middle queues, frees through C2 then C3; verifies no bypass.
  - `chain_concurrency_to_ratelimit_resumes_with_held_queue` — `[C, RateLimit]` resume task carries the C holder paired with the request_id used to create the holder row.
- [x] Existing suites pass: `holder_leak_tests`, `floating_concurrency_tests`, `concurrent_grant_race_tests`, `job_store_shard_enqueue_tests`, `job_store_shard_restart_tests`, `job_store_shard_cancel_tests`, `concurrency_tests`, `concurrency_requester_counter_tests`, `codec_tests`, `job_store_shard_retry_tests`, `job_store_shard_import_tests`, `job_store_shard_concurrency_tests` (26 + 3 new).

### DIFF
--- a/schema/internal_storage.fbs
+++ b/schema/internal_storage.fbs
@@ -129,7 +129,6 @@ table ResumeAfterConcurrencyGrant {
   task_group: string;
   held_queues: [HeldQueue];
   limit_index: uint32;
-  start_at_ms: int64;
   priority: ubyte;
 }
 
--- a/schema/internal_storage.fbs
+++ b/schema/internal_storage.fbs
@@ -31,6 +31,15 @@ table KeyValuePair {
   value: string;
 }
 
+// Pair of (queue, task_id) identifying a concurrency holder created during the
+// limit chain. Each chain accumulates one HeldQueue per granted limit; the
+// task_id is the request_id that created the holder, used at release time to
+// look up `concurrency_holder_key(tenant, queue, task_id)`.
+table HeldQueue {
+  queue: string;
+  task_id: string;
+}
+
 table RetryPolicy {
   retry_count: uint32;
   initial_interval_ms: int64;
@@ -74,7 +83,7 @@ table RunAttempt {
   job_id: string;
   attempt_number: uint32;
   relative_attempt_number: uint32;
-  held_queues: [string];
+  held_queues: [HeldQueue];
   task_group: string;
 }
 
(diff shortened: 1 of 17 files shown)


kirinrastogi:
this approach isn't feasible because it get JobInfo in the hot path. the limits array itself needs to be persisted to work around that

PR #319, opened 2026-05-19
Avoid thundering-herd for GetClusterInfo by blocking on single Future

udnay:
- Collapse concurrent GetClusterInfo callers onto a single in-flight coordinator fetch using futures::future::Shared, eliminating thundering-herd load on the coordination layer (k8s/etcd) when many routing clients refresh at once.
  - Extract the build logic into SiloService::fetch_cluster_info so it can be wrapped in an owned 'static + Send future stored in a new cluster_info_inflight slot; the slot is cleared when the future resolves so subsequent calls still get
  fresh data.
  - Preserves all existing semantics (5s timeout, cached-fallback on coordinator unavailability, Status::unavailable when no cache exists).

  - cargo test --test cluster_info_cache_tests — all 4 tests pass
  - New get_cluster_info_dedupes_concurrent_callers test: fires 50 concurrent callers against a coordinator with 300ms artificial latency and asserts:
    - get_shard_owner_map is called exactly once (not 50×)
    - get_members is called exactly once
    - All 50 callers receive identical response   - A follow-up call after the herd resolves produces a fresh fetch (slot cleared)
  - Existing timeout / cache-fallback / recovery tests still pass

---

> [!NOTE]
> **Medium Risk**
> Touches `GetClusterInfo` routing/topology RPC behavior and introduces shared in-flight async state; mistakes could cause stale responses, deadlocks, or more frequent `Unavailable` errors under load, though semantics and timeouts are largely preserved.
> 
> **Overview**
> **`GetClusterInfo` now collapses concurrent requests into a single coordinator fetch** by storing a shared in-flight future (`Shared<BoxFuture<...>>`) in `SiloService` and having all callers await it.
> 
> The coordinator-fetch + timeout + cached-fallback logic is extracted into `SiloService::fetch_cluster_info`, and the in-flight slot is cleared once the future resolves so subsequent calls fetch fresh data.
> 
> Tests extend the mock coordinator to track call counts/latency and add `get_cluster_info_dedupes_concurrent_callers`, asserting 50 concurrent callers produce exactly one `get_shard_owner_map` + one `get_members` call and that a post-herd call triggers a new fetch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8674ccd8d4bef316346810626f2c21ab421aa5cf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -4,6 +4,8 @@ use std::sync::Arc;
 use std::sync::LazyLock;
 use std::time::Duration;
 
+use futures::FutureExt;
+use futures::future::{BoxFuture, Shared};
 use rand::seq::SliceRandom;
 use tokio::net::TcpListener;
 use tokio::sync::{Mutex, broadcast};
@@ -203,6 +205,12 @@ fn proto_metadata_to_domain(metadata: HashMap<String, String>) -> Option<Vec<(St
 /// server) don't take routing clients down with them.
 const CLUSTER_INFO_TIMEOUT: Duration = Duration::from_secs(5);
 
+/// Shared in-flight future returned to concurrent `GetClusterInfo` callers
+/// so a thundering herd collapses onto a single coordinator fetch. The
+/// `Ok` variant is the response to return; the `Err` variant is a status
+/// message to wrap in `Status::unavailable`.
+type ClusterInfoFuture = Shared<BoxFuture<'static, Result<GetClusterInfoResponse, String>>>;
+
 /// gRPC service implementation backed by a `ShardFactory`.
 #[derive(Clone)]
 pub struct SiloService {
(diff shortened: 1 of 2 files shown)


udnay:

- #319 ◀

src/server.rs
```diff
@@ -213,6 +221,9 @@ pub struct SiloService {
     /// Last successful `GetClusterInfo` response, served as a fallback when
     /// the coordination layer is slow or unreachable.
     cluster_info_cache: Arc<Mutex<Option<GetClusterInfoResponse>>>,
+    /// In-flight `GetClusterInfo` future, shared across concurrent callers
+    /// so only one fetch hits the coordinator at a time.
+    cluster_info_inflight: Arc<Mutex<Option<ClusterInfoFuture>>>,
```
kirinrastogi, line 226: I don't know if I like this approach because it increases the latency unnecessarily. The clients have a retry budget, and I  think we should switch this method to no longer be consistent.

What do you think about a background task that writes to a cache once every 5s, and all grpc calls are populated from the cache? And on startup synchronously block until the cache is populated.

kirinrastogi (approved):

PR #323, opened 2026-05-21
Unify limit granting flow

kirinrastogi:
## Summary

- Unifies the immediate-enqueue and `process_grants` code paths so multi-limit chains follow the same canonical walker (`enqueue_limit_task_at_index`) in every continuation context (fresh enqueue, retry, reimport, future-scheduled `RequestTicket`, deferred concurrency request resume, rate-limit follow-up).
- Refactors `LimitTaskParams` to split the previously-overloaded `start_at_ms` into two semantically distinct fields: `scheduled_at_ms` (drives the future-scheduling decision; persisted on `EnqueueTask`) and `task_key_start_ms` (the time component baked into every `task_key`/`concurrency_request_key` this walk writes). The chain resumer is the one call site where these two diverge.
- Fixes a holder leak in multi-limit chains where `process_grants`' resumer re-emitted the resumed `RunAttempt` at the *same* `task_key` the original enqueue had written (and then deleted in-batch). If the broker had ack-deleted that key from any prior visibility, every subsequent write was silently suppressed by the broker tombstone, stranding the holders the resumer had just granted. Mirrors the long-standing precedent in `handle_request_ticket` (dequeue.rs) and the `project_broker_tombstone_chain_continuation` memory.
- Fixes three held-queue leak vectors uncovered by review of the unified flow (see "Holder-leak fixes" below).
- Adds `task_id` and `held_queues` columns to the `tasks` SQL view so future holder-leak audits can directly compare leaked-holder `task_id`s against the `RunAttempt` tasks in the DB.

## Why tombstones

`TaskBroker::ack_tombstones` is silo's own in-memory map (not slatedb's LSM tombstones). When a worker durably ack-deletes a `task_key`, it lands in this map keyed by the full key bytes; subsequent scans that observe the same key are skipped *and* the tombstone's generation is refreshed, so a key that keeps reappearing in the DB stays suppressed indefinitely. The chain resumer was the only chain-continuation entrypoint not following the rule "write follow-up tasks at a fresh `task_key`" — closing that gap is what the `task_key_start_ms` split does.

## What changed (scheduled_at_ms vs task_key_start_ms)

| Call site | `scheduled_at_ms` | `task_key_start_ms` |
|---|---|---|
| Fresh enqueue (immediate / future) | user's `start_at_ms` | same |
| Retry after error | `next_time` (backoff) | same (`attempt_number+1` keeps the key unique) |
| Reimport | request `start_at_ms` | same (fresh `attempt_number`) |
| `ShardChainResumer::resume_chain` | `params.start_at_ms` (honest) | **`now_ms`** (different — dodges the tombstone) |
| `handle_request_ticket` / `handle_check_rate_limit` chain continuations | `now_ms` | `now_ms` |

## Holder-leak fixes (commit `2727466`)

Three structural leaks the unified flow exposed — all in paths that didn't exist before this PR (or that the prior chain code couldn't reach):

1. **`handle_request_ticket` terminal/missing-job path** (`dequeue.rs`). When the persisted job status is `Succeeded`/`Failed`/`Cancelled` or the status row is missing/unreadable, the code deletes the ticket and returns. Post-PR a future-scheduled multi-concurrency chain can reach this path with `stored_held_queues = [A, …]`; if the job has since terminated, those holders are silently orphaned. Now mirrors `handle_check_rate_limit`'s missing-`JobInfo` branches: batches the holder deletes and queues them for post-commit `atomic_release`.
2. **`cancel_job` `Task::RequestTicket` arm** (`cancel.rs`). Pre-PR the comment "deleting the task is sufficient" was true because tickets had no `held_queues`. Post-PR a `FutureRequestTaskWritten` ticket can carry prior grants; cancel now destructures `held_queues`/`task_id`, deletes each holder, and feeds them into the existing post-commit `atomic_release` path.
3. **`handle_request_ticket` / `handle_check_rate_limit` rollback on handler `Err`** (`dequeue.rs`). The per-iteration `state.grants_to_rollback` was only merged into the outer accumulator after all handlers in the iteration succeeded — and the outer rollback only ran on `write_with_options` failure. So an `Err` returned from inside a handler (e.g. an `enqueue_limit_task_at_index` failure after a successful `try_reserve`) bypassed *both* points, leaving the in-memory reservation orphaned. Per-iteration handler `Err` now drains `state.grants_to_rollback` and rolls back each grant before requeueing claimed entries and propagating. `state.holder_releases` is intentionally *not* drained — its batched deletes never committed, so the in-memory holders must persist.

Plus minor follow-ups from the same review:

- Tightened `two_concurrency_limits_chain_writes_new_request_when_b_full`: A capacity raised from 2 to 3 so all three jobs deterministically reach the `limit_index=1` deferred-B state; the final assertion is now load-bearing (was `let _ = saw_resumable_state`).
- Added `debug_assert_eq!(scheduled_at_ms, task_key_start_ms)` on the future-`RequestTicket` branch so a future call site that ever passes them apart trips immediately instead of silently desyncing the persisted `start_time_ms` from the `task_key`.
- Added a code comment near the grant scanner's `start_time_ms > now_ms` break explaining why the early termination is safe given the request-key encoding.

## Test plan

- [x] `cargo test` across holder_leak / floating / concurrent_grant_race / concurrency / job_store_shard_concurrency / dequeue / enqueue / retry / rate_limit / import — 209 tests, 0 failures.
- [x] Added regression test `resume_chain_writes_run_attempt_at_now_not_original_start_at_ms` (fails on the prior code; passes with the fix).
- [x] Added regression test `cancel_future_request_ticket_releases_held_queues` (future-schedules `[A free, B full]`, asserts the future `RequestTicket` carries `held_queues=[A]`, cancels, then asserts A's holder count is zero).
- [x] `cargo fmt` / `cargo check --all-targets` / `cargo clippy --all-targets` clean on the touched files.
- [ ] Could not locally reproduce the original leak (4 silo-bench runs at the prior agent's exact config, all clean). The brief's repro rate was ~2/3 — my env didn't trigger it. The regression tests pin down the *invariants* the fix establishes; the leak vector itself (slatedb scan-vs-batch ordering visibility) is too timing-sensitive to exercise without fault injection.
- [ ] CI: https://github.com/gadget-inc/silo/actions/runs/26244290782

## Caveats

- Splitting `LimitTaskParams` touches six files (handle_enqueue signature change). All callers updated; no behavioural change vs the one-line fix in `901df7c`.
- The `query.rs` change (new `task_id`/`held_queues` columns on `tasks` view) is incidental — it's the audit tool I wanted while debugging. Happy to drop it if you'd rather keep this PR scoped tighter.
- Persisting `limits: Vec<Limit>` on every deferred concurrency request and every future `RequestTicket` is a storage-vs-fetch tradeoff: it saves a `JobInfo` round-trip on resume, at the cost of duplicating the list once per deferred record. For high-fanout jobs with many limits this inflates each request materially.


kirinrastogi:
closing for https://github.com/gadget-inc/silo/pull/326

PR #330, opened 2026-05-28
Introduce terminal-job expiration helpers and settings

kirinrastogi:
## Summary

These helpers expire jobs transactionally, by re-writing them and their metadata with new expire_ts.

PR 2/4 in the job-retention stack split. Stacks on #329 (`kirin/job_retention/counter_reconciliation_settings`).

Adds the plumbing the cancel / import / completion paths need to apply a SlateDB row TTL on terminal job records. **No production caller is wired up yet** — that lands in the next two PRs in this stack.

What this PR adds:

- `DatabaseTemplate` / `DatabaseConfig` gain `completed_job_expire_s` and `terminal_job_expire_s`, both `Option<u64>` and defaulted to `None`.
- `JobStoreShard` carries both values through `OpenShardOptions`.
- New `compute_terminal_expire_ts` resolver (and its `JobStoreShard` wrapper) computes `now_ms + seconds * 1000` with saturating arithmetic so an absurd operator value cannot wrap to a negative `expire_ts`.
- `WriteBatcher::put_with_expire` is added to the trait with **no default impl** — every implementor must thread the TTL through explicitly. A silent fallback that dropped the TTL would produce data that never expires with no compile-time signal, so the trait forces each writer to decide. Real overrides on `DbWriteBatcher` and `TxnWriter` route through `slatedb::config::PutOptions { ttl: Ttl::ExpireAt(ts) }`. `InstrumentedDbTransaction` gains a `put_with_options` shim.
- `set_job_status_with_index_opts` / `write_job_status_with_index_opts` accept an optional `expire_ts` and tag JOB_STATUS + IDX_STATUS_TIME via `put_with_expire`.
- `expire_terminal_job_records` re-puts JOB_INFO, IDX_METADATA pairs, every prior ATTEMPT row, and JOB_CANCELLED (if present) with the row TTL.

Unused-yet items carry a narrow `#[allow(dead_code)]` pointing at the follow-up PRs; those allows go away when the callers land.

## Test plan
- [x] `cargo fmt`
- [x] `cargo clippy --workspace --lib --bins --all-features -- -D warnings`
- [x] `cargo check --all-targets`
- [x] Inline unit tests for `compute_terminal_expire_ts` (Succeeded vs other terminal, missing config, u64→i64 overflow saturation, `now_ms` near `i64::MAX`).
- [x] `cargo test --test expire_helper_tests` — 5 passed (TOML round-trip + shard-open smoke).
- [x] `cargo test --test config_and_db_tests` — 35 passed, no regressions.
- [ ] End-to-end coverage of `expire_terminal_job_records` itself comes via the integration tests in the follow-up cancel/import (PR 3/4) and completion/failure (PR 4/4) PRs that wire the helper to its real callers.

### DIFF
--- a/benches/bench_helpers.rs
+++ b/benches/bench_helpers.rs
@@ -614,6 +614,8 @@ pub async fn clone_golden_shard(
             ),
             counter_reconciliation_seconds: None,
             hydrate_all_at_startup: false,
+            completed_job_expire_s: None,
+            terminal_job_expire_s: None,
         },
         ShardRange::full(),
     )
--- a/src/factory.rs
+++ b/src/factory.rs
@@ -193,6 +193,8 @@ impl ShardFactory {
                         ),
                         counter_reconciliation_seconds: template.counter_reconciliation_seconds,
                         hydrate_all_at_startup: template.hydrate_all_at_startup,
+                        completed_job_expire_s: template.completed_job_expire_s,
+                        terminal_job_expire_s: template.terminal_job_expire_s,
                     },
                     range.clone(),
                 )
--- a/src/instrumented_db.rs
+++ b/src/instrumented_db.rs
@@ -20,7 +20,7 @@ use std::ops::RangeBounds;
 use std::sync::Arc;
 
 use slatedb::bytes::Bytes;
-use slatedb::config::{FlushOptions, ScanOptions, WriteOptions};
+use slatedb::config::{FlushOptions, PutOptions, ScanOptions, WriteOptions};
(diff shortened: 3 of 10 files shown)


udnay (approved):
LGTM,

This still means that any updated expiration timestamp needs to make it all the way to the bottom sorted run level right to get dropped?

kirinrastogi:
Yes. To enforce that we can add a ticker that submits CompactionSpecs periodically.

PR #338, opened 2026-06-01
Add periodic leaked-holder reaper

kirinrastogi:
## Problem

A tenant's concurrency queue can wedge permanently due to **orphaned concurrency holders**. A holder (KV prefix `0x09`) is granted the instant a `RunAttempt` task is created, tied to a `task_id`, and is normally released on `report_attempt_outcome`, `cancel_job`, or lease-expiry reaping. When a worker pulls a `RunAttempt` into its in-flight buffer and drops the dequeue/lease future **before a lease is durably written**, the holder is left with no lease and no task.

Nothing iterates holders to recover this:
- the lease reaper iterates *leases* — a holder with no lease is invisible to it;
- `reconcile_pending_requests` iterates *requests* and never inspects holders;
- `purge_orphaned_holders_for_task` only fires on a late `report_outcome` for that exact `task_id`, which never arrives when the worker is gone.

So the holder stays pegged until ghost holders fill the cap, `try_reserve`/`process_grants` see the queue as full, and the requester backlog starves forever.

## Change

A periodic, per-shard reaper that scans holders and releases any orphaned one — no lease for its `task_id` and a grant older than a configurable grace window — via the **full** release path (durable delete + `atomic_release` + `request_grant`), never a bare KV delete.

- **`reap_orphaned_holders(grace_ms, max_per_tick)`** (`src/job_store_shard/lease.rs`): single `0x09`-prefix scan per shard, split-aware, scan-then-point-get to avoid false positives (a lease written concurrently is observed and the holder skipped). Grant wakeups and warn logs coalesce per queue; a per-tick cap bounds the durable batch so a large backlog drains over several ticks.
- **Scheduling** (`src/server.rs`): a second 1s reaper task next to the lease reaper, decoupled via its own shutdown receiver so a backpressured holder scan can't stall lease reaping (mirrors the decoupled metrics-scrape task).
- **Settings** (`src/settings.rs`): `orphaned_holder_grace_ms` (default 60s — 6× `DEFAULT_LEASE_MS`) and `orphaned_holder_max_reap_per_tick` (default 1000).
- **Metrics** (`src/metrics.rs`): `silo_holder_reaper_{duration_seconds,scans_total,orphans_reaped_total,errors_total}`, mirroring the lease-reaper set.

## Bounding / fan-out

One scan per shard per second, sequential over shards on a single task (no overlap). Within a scan, point-gets and releases are sequential — no spawn/join. `request_grant` coalesces by `(tenant, queue)` into one pending-grant entry with an accumulated count, so a queue with hundreds of orphans triggers one `process_grants` pass, not one per holder. The per-tick cap bounds the durable write under a pathological backlog.

## Tests

`tests/holder_leak_tests.rs` — releases a stranded holder; skips holders with a lease / younger than grace / for an out-of-range tenant; respects the per-tick cap (3 orphans, cap 2 → reaps 2 then 1).

## Verification

- `cargo build` clean, `cargo fmt --check` clean
- `cargo clippy` — no new warnings in changed source
- `cargo test --test holder_leak_tests` — 15/15 pass

### DIFF
--- a/src/job_store_shard/lease.rs
+++ b/src/job_store_shard/lease.rs
@@ -5,8 +5,8 @@ use slatedb::config::WriteOptions;
 use uuid::Uuid;
 
 use crate::codec::{
-    decode_floating_limit_state, decode_lease, encode_attempt, encode_floating_limit_state,
-    encode_lease,
+    decode_floating_limit_state, decode_holder_granted_at_ms, decode_lease, encode_attempt,
+    encode_floating_limit_state, encode_lease,
 };
 use crate::dst_events::{self, DstEvent};
 use crate::job::{FloatingLimitState, JobStatus, JobView};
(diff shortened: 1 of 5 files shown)


kirinrastogi:
it's hard to tell which holders are safe to delete

PR #346, opened 2026-06-02
Add "Drop holders & run attempts" tenant admin action

kirinrastogi:
image

clients see this:

```
2026-06-02T17:44:13.848492Z  INFO silo::routing_client: Routing error detected, refreshing topology
2026-06-02T17:44:13.864159Z  INFO silo::routing_client: Routing error detected, refreshing topology
2026-06-02T17:44:13.867352Z  WARN silo_bench: Worker failed to report outcome after retry worker_id=bench-worker-3 error=status: NotFound, message: "lease not found", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Tue, 02 Jun 2026 17:44:13 GMT", "content-length": "0"} }
2026-06-02T17:44:13.888029Z  WARN silo_bench: Worker failed to report outcome after retry worker_id=bench-worker-9 error=status: NotFound, message: "lease not found", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Tue, 02 Jun 2026 17:44:13 GMT", "content-length": "0"} }
2026-06-02T17:44:14.299887Z  INFO silo_bench: Benchmark progress elapsed_secs=67.0960995 completed=5202 completed_rate_per_sec=75.871104477381 enqueued=5222 enqueue_rate_per_sec=77.86771248994364 empty_polls=9595 refreshes=0
   67.1s         5202         75.9         5222         77.9         9595            0
```

there is a confirm:

image

## What

Adds a red, confirmation-gated **"Drop holders & run attempts"** button to the tenant view (`/tenant`) that forcibly clears a tenant's stuck concurrency state on a single shard — for unsticking a tenant whose concurrency holders have leaked.

The browser confirmation uses HTMX's `hx-confirm` (native `window.confirm()` fired before the request is sent), matching the existing red destructive-button style from `job.html`.

## How it works

New shard method `drop_tenant_holders_and_runattempts(tenant)` is **intentionally destructive and non-transactional**, mirroring the bulk-delete pattern in `cleanup.rs`:

1. Read-only scans collect: the tenant's holders (tenant-prefixed), queued `RunAttempt` tasks (TASK prefix, filtered by the tenant inside the task), and running/leased `RunAttempt`s (LEASE prefix, via `decode_lease().to_task()`).
2. All three are deleted **together** via plain chunked `WriteBatch`es (`db.write`, no transaction).
3. Post-delete clean release: `atomic_release` + `request_grant` per holder, and broker eviction for deleted tasks.

### Why not a transaction?

This has to run on a **live env with no downtime**. Run-attempt tasks and leases aren't tenant-keyed, so finding them requires scanning the *entire* TASK + LEASE keyspaces. Under `SerializableSnapshot` that pulls every other tenant's concurrent job activity into the conflict set, and `retry_on_txn_conflict` is bounded (5 tries, ~150ms) — so on a busy shard it reliably fails with `TransactionConflict`, precisely when you need it. The batched-delete approach always commits regardless of concurrent activity.

Trade-offs (all benign for an emergency tool): **point-in-time / best-effort** (work created after the scan isn't caught — re-run; the op is idempotent) and **partial progress is safe** (a holder without its lease self-heals on the worker's next heartbeat).

### Worker safety

If a worker reports an outcome for a run attempt we dropped, it's **not** silently dropped: `report_attempt_outcome` checks the lease first and returns `LeaseNotFound` before any mutation, which the server maps to `NOT_FOUND` and even self-heals orphaned holders (`purge_orphaned_holders_for_task`). This is the same path normal lease expiry/reaping already produces, so workers handle it.

## Scope

Single shard (resolved from the tenant's jobs; the response notes if the tenant spans multiple shards). Historical `ATTEMPT` records and concurrency *requests* are deliberately left untouched.

## Wiring

- `proto/silo.proto` — new `DropTenantState` RPC + request/response messages
- `src/server.rs`, `src/cluster_client.rs` — gRPC server impl + local/remote client routing
- `src/webui.rs`, `templates/tenant.html` — `/tenant/drop-state` route, handler, and button
- `src/job_store_shard/drop_tenant.rs` *(new)* — the drop method

## Tests

- `tests/job_store_shard_drop_tenant_tests.rs` *(new)* — drops a queued + a leased run attempt plus holders for one tenant, asserts durable + in-memory state cleared and a second tenant untouched; plus a no-op-for-empty-tenant case.
- Added the new method to the mock `Silo` impl in `tests/routing_client_tests.rs`.

`cargo fmt`, `cargo check --all-targets`, and clippy are clean for the changed code; adjacent suites (cancel, concurrency, holder_leak, webui, routing_client) pass.


src/job_store_shard/drop_tenant_holders.rs
```diff
@@ -0,0 +1,181 @@
+//! Tenant-wide forced cleanup of concurrency holders and in-flight run attempts.
```
udnay, line 1: This file name is pretty scary, can we give it a cuddlier name?

udnay (approved):
Chatted about this IRL

PR #347, opened 2026-06-03
Fix granted-RunAttempt back-of-queue leak via task_key epoch_ms

kirinrastogi:
## Problem

The wedge incident: a tenant wedged — every queue pinned at max holders, **0 running jobs**, holders 17–30h old and surviving restarts (durable, not in-memory ghosts).

When the grant scanner resumes a waiting job's limit-chain, the terminal `RunAttempt` was stamped into the broker at `now_ms` (grant time) instead of the job's original `scheduled_at_ms`, in order to dodge the broker **ack-tombstone** of the just-deleted predecessor `task_key`. Because it sorted at grant time, the granted RunAttempt landed at the **back** of the broker's `start_time_ms`-ascending order — behind the tenant's wall of old, ungrantable `RequestTicket`s.

The broker doesn't scan straight to dispatch: it refills an in-memory buffer (`target_buffer`, ordered by `start_time_ms` ascending) by re-scanning the task keyspace **from the front every generation**, and dispatches from the front of that buffer. So this wasn't a one-shot "capped scan stopped short" — it was steady-state starvation. The old `RequestTicket`s sit at the front (old `start_time`s) and never leave: a ticket whose queue is full fails to reserve and is *left in place* (`ack_release`, not deleted/tombstoned), so every generation re-buffers the same front-most wall. The broker spent its cycles endlessly re-dispatching those tickets — work that could never complete, because the slots they needed were held by the grant-time RunAttempt itself stranded *behind* the wall. With more than `target_buffer` live tickets ahead of it, the RunAttempt never entered the buffer → never ran → never released its concurrency holders → the wall stayed ungrantable → self-reinforcing wedge.

## Fix

Decouple the tombstone dodge from broker ordering by adding a trailing `epoch_ms` field to `task_key`:

```
(task_group, start_time_ms, priority, job_id, attempt, epoch_ms)
              ^^ = scheduled_at_ms (logical)            ^^ = now_ms-style dodge
```

- `start_time_ms` always equals the logical `scheduled_at_ms` (governs broker ordering) — so a granted RunAttempt sorts by its true schedule and stays reachable.
- `epoch_ms` is a **write-only** disambiguator (`now_ms`, or `now_ms.max(parent_epoch_ms + 1)` on a rewrite). It sorts last, so it never perturbs cross-job ordering or the scan's early-break.
- The deleted predecessor's tombstone stays intact for its real purpose (suppressing stale-scan re-inserts of a just-deleted interim).

### Who needs `epoch_ms` for lookups? Nobody — it's write-only
- The broker / dequeue / delete / ack path is already key-agnostic (it parses discovered keys and carries raw bytes).
- The four "reconstruct-exact-key + `get`" control-plane sites (expedite, cancel, lease_task, import) now use `helpers::find_task_by_identity` — a single-result prefix scan on the 5 identity fields (`keys::task_key_lookup_prefix`), omitting `epoch_ms`. Only one chain task is live per `(job, attempt)`, so the scan returns ≤1.

## Changes
- **`keys.rs`**: `epoch_ms` on `task_key` / `parse_task_key` / `ParsedTaskKey`; new `task_key_lookup_prefix`.
- **Write path**: replace `LimitTaskParams.task_key_start_ms` with `task_key_epoch_ms`; thread through `put_task`, `handle_enqueue`, `walk_limit_chain`, `limit_chain::resume_chain`, `handle_request_ticket`, `handle_check_rate_limit`, `schedule_rate_limit_retry`, and fresh-chain / refresh / restart / range-bound sites.
- **Read path**: the four point-lookup sites switch to `find_task_by_identity` prefix scans.

**Pre-launch:** old-layout task keys are flushed on deploy — no migration / decode fallback.

## Tests
- `granted_run_attempt_keeps_original_start_time` (concurrency suite) — **fails on `main`**, passes here. Verified the pre-fix key carried grant time (~1h after the original schedule).
- Rewrote `holder_leak` resume test → `resume_chain_writes_run_attempt_at_original_start_with_fresh_epoch` (asserts original start + a fresh epoch). The old test asserted the now-fixed buggy behavior.
- Fixed `dequeue_ignores_recently_acked_task_keys` to reinject the real key (it hardcoded `epoch=0`).
- Extended `keys_tests` for epoch ordering + the lookup prefix.

`cargo fmt`, `cargo check --tests`, and `cargo clippy --lib` are clean. Full suite green except `k8s_coordination_tests` (pre-existing environmental: `auth exec` binary absent).

### DIFF
--- a/src/job_store_shard/helpers.rs
+++ b/src/job_store_shard/helpers.rs
@@ -8,7 +8,7 @@ use crate::codec::encode_task;
 use crate::instrumented_db::{InstrumentedDb, InstrumentedDbIterator, InstrumentedDbTransaction};
 use crate::job::JobStatus;
 use crate::job_store_shard::JobStoreShardError;
-use crate::keys::{end_bound, task_key};
+use crate::keys::{end_bound, task_key, task_key_lookup_prefix};
 use crate::task::Task;
 
 /// A trait that abstracts over SlateDB's two ways of writing in groups: `WriteBatch` and `DbTransaction`.
@@ -239,6 +239,12 @@ pub fn now_epoch_ms() -> i64 {
 }
 
 /// Encode and write a task to a batch or transaction at the standard task key location.
+///
+/// `time_ms` is the logical start time (governs broker ordering); `epoch_ms` is
+/// the write-time disambiguator that lets a chain rewrite dodge the broker
+/// ack-tombstone of a just-deleted predecessor without perturbing ordering. See
+/// [`crate::keys::task_key`].
+#[allow(clippy::too_many_arguments)]
 pub(crate) fn put_task<W: WriteBatcher>(
     writer: &mut W,
     task_group: &str,
@@ -246,14 +252,42 @@ pub(crate) fn put_task<W: WriteBatcher>(
     priority: u8,
     job_id: &str,
     attempt: u32,
+    epoch_ms: i64,
     task: &Task,
 ) -> Result<(), JobStoreShardError> {
     let task_value = encode_task(task);
     writer.put(
-        task_key(task_group, time_ms, priority, job_id, attempt),
+        task_key(task_group, time_ms, priority, job_id, attempt, epoch_ms),
         &task_value,
     )?;
     Ok(())
+}
+
+/// Find the single live task identified by
+/// `(task_group, start_time_ms, priority, job_id, attempt)`, ignoring the
+/// trailing write-time `epoch_ms`. Returns the task's full key bytes and encoded
+/// value, or `None` if absent.
+///
+/// Status-driven point lookups (lease, cancel, expedite, reimport) know a task's
+/// identity from `JobStatus` but not its `epoch_ms`, which is a write-only
+/// disambiguator (see [`crate::keys::task_key`]). Only one chain task is ever
+/// live per `(job_id, attempt)` — a rewrite deletes the old key and writes the
+/// new one in the same batch — so this prefix scan returns at most one row.
+pub(crate) async fn find_task_by_identity(
+    txn: &InstrumentedDbTransaction,
+    task_group: &str,
+    start_time_ms: i64,
+    priority: u8,
+    job_id: &str,
+    attempt: u32,
+) -> Result<Option<(Vec<u8>, Bytes)>, JobStoreShardError> {
+    let prefix = task_key_lookup_prefix(task_group, start_time_ms, priority, job_id, attempt);
+    let end = end_bound(&prefix);
+    let mut iter = txn.scan::<Vec<u8>, _>(prefix..end).await?;
+    match iter.next().await? {
+        Some(kv) => Ok(Some((kv.key.to_vec(), kv.value))),
+        None => Ok(None),
+    }
 }
 
 /// Retry an operation that may fail with a SlateDB transaction conflict.
--- a/src/job_store_shard/import.rs
+++ b/src/job_store_shard/import.rs
@@ -14,13 +14,13 @@ use crate::fb::silo::fb;
 use crate::job::{JobInfo, JobStatus, JobStatusKind, JobView, Limit};
 use crate::job_attempt::{AttemptStatus, JobAttempt};
 use crate::job_store_shard::helpers::{
-    TxnWriter, decode_job_status_owned, now_epoch_ms, put_with_optional_expire,
-    retry_on_txn_conflict,
+    TxnWriter, decode_job_status_owned, find_task_by_identity, now_epoch_ms,
+    put_with_optional_expire, retry_on_txn_conflict,
 };
 use crate::job_store_shard::{JobStoreShard, JobStoreShardError, LimitTaskParams};
 use crate::keys::{
     attempt_key, attempt_prefix, concurrency_holder_key, end_bound, idx_metadata_key,
-    job_cancelled_key, job_info_key, job_status_key, task_key,
+    job_cancelled_key, job_info_key, job_status_key,
 };
 use crate::retry::{RetryPolicy, retries_exhausted};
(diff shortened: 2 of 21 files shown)


src/job_store_shard/helpers.rs
```diff
@@ -239,23 +239,57 @@ pub fn now_epoch_ms() -> i64 {
...
+    task_group: &str,
+    start_time_ms: i64,
+    priority: u8,
+    job_id: &str,
+    attempt: u32,
+) -> Result<Option<(Vec<u8>, Bytes)>, JobStoreShardError> {
+    let prefix = task_key_lookup_prefix(task_group, start_time_ms, priority, job_id, attempt);
+    let end = end_bound(&prefix);
+    let mut iter = txn.scan::<Vec<u8>, _>(prefix..end).await?;
+    match iter.next().await? {
```
udnay, line 287: We only grab the first one? The vector is of size 1 here always?

src/job_store_shard/import.rs
```diff
@@ -287,11 +287,11 @@ impl JobStoreShard {
...
                         priority: params.priority,
-                        // Fresh import → fresh chain. Scheduled and task_key
-                        // time coincide; this attempt_number is unseen so no
-                        // tombstone can collide.
+                        // Fresh import → fresh chain. The task_key starts at the
+                        // scheduled time; this attempt_number is unseen so the
+                        // epoch is just the write time.
                         scheduled_at_ms: effective_start_at_ms,
-                        task_key_start_ms: effective_start_at_ms,
+                        task_key_epoch_ms: now_ms,
```
udnay, line 294: What happens if `effective_start_at_ms` is at time `t` and `now_ms` is time `t+10` so we have
```
LimitTaskParams {
...
  scheduled_at_ms: effective_start_at_ms (t),
  task_key_epoch_ms: now_ms (t+10),
...
}
```
 but when we write a `LimitTask` in dequeue we set it to `t+1`, that's for a `CheckRateLimit` task so maybe these are different flows. Why not keep things consistent throughout though? Either `parent +1` or `now`

I'm curious if this would have any impact on how we look up the tasks.

src/job_store_shard/lease_task.rs
```diff
@@ -137,26 +135,35 @@ impl JobStoreShard {
...
-                    Some(raw) => (zero_key, raw),
-                    None => {
-                        return Err(JobStoreShardError::JobNotLeaseable(JobNotLeaseableError {
-                            job_id: id.to_string(),
-                            status: status.kind,
-                            reason: "job has no pending task in queue".to_string(),
-                        }));
-                    }
-                }
+                find_task_by_identity(&txn, &task_group, 0, priority, id, attempt_number).await?
```
udnay, line 156: When would the start time ever be 0?

src/job_store_shard/rate_limit.rs
```diff
@@ -31,11 +31,16 @@ impl JobStoreShard {
...
+        parent_epoch_ms: i64,
         task_group: &str,
     ) -> Result<(), JobStoreShardError> {
-        let task_key_start_ms =
-            rate_limit_retry_task_key_start_ms(retry_at_ms, parent_start_time_ms);
+        // The retry's task_key keeps `retry_at_ms` as its start time (the broker
+        // must dispatch it after the backoff). `handle_check_rate_limit`
+        // ack-deleted the parent CheckRateLimit, installing a broker tombstone at
+        // its task_key; every other key component is identical to the parent's,
+        // so we dodge it with an `epoch_ms` strictly past the parent's. This
```
udnay: Is this statement true, we're just using the parent's epoch not increasing it here.

udnay: Ah, it's buried in a function here.

udnay (approved):
A couple of questions but this looks fine

src/job_store_shard/helpers.rs
kirinrastogi: Yes, I don't think this can use `get` because we don't know the epoch_ms.

src/job_store_shard/lease_task.rs
kirinrastogi: immediate enqueue path

src/job_store_shard/helpers.rs
udnay: Not using `get` makes sense. It's the epoch_ms marching forward the means epoch_ms `t+1` would be after `t` if for some reason we have 2 tasks there. Not sure if that's possible or if it really matters.

src/job_store_shard/lease_task.rs
udnay: 😱

kirinrastogi: I'm going to update this so that they get enqueued at the `now()` when enqueued. But we still need to handle the 0 case I think.

kirinrastogi: We don't need to worry about this change because we enqueue with a ts for immediate enqueue. But I think the grpc api allows any ts for scheduled time.

src/job_store_shard/import.rs
kirinrastogi: I think the `t+1` you mean is written to the `epoch_ms` field, which isn't necessary for lookups. 

When the effective_start_at_ms is 10 behind now_ms, the RequestTicket will be enqueued at that timestamp, and be elligible for dequeueing immediately.

src/job_store_shard/helpers.rs
kirinrastogi: Only one chain task is ever live per `(job_id, attempt)` — a rewrite deletes the old key and writes the new one in the same batch — so this prefix scan returns at most one row

PR #349, opened 2026-06-03
Make grant-scanner throughput clamps configurable

kirinrastogi:
When we have a huge job backlog we see that the grant scanner can't fill the task broker fast enough, but there are still outstanding jobs:

image

We want to be able to tune these values in config.

## What

The grant scanner (`ConcurrencyManager::process_grants`) bounds its per-pass throughput with two hardcoded constants. This moves both into the TOML database settings so they can be tuned per deployment without a rebuild. **Defaults are unchanged.**

| Setting (TOML, under `[database]`) | Type | Default | Replaces |
|---|---|---|---|
| `grant_scanner_batch_size` | `usize` | `256` | `MAX_GRANTS_PER_PASS` — max grants materialized / status-checked / committed in a single pass |
| `grant_scanner_buffer_size` | `usize` | `64` | `STATUS_LOOKUP_CONCURRENCY` — max in-flight `db.get(job_status_key)` lookups via `.buffered(N)` |

## How

Mirrors the existing `hydrate_all_at_startup` numeric-setting pattern:

- **`src/settings.rs`** — `pub const` defaults + `default_*` fns; fields added to both `DatabaseTemplate` and `DatabaseConfig` and their `Default` impls.
- **`src/job_store_shard/mod.rs`** — fields on `OpenShardOptions`, threaded into `ConcurrencyManager::new`; populated from `cfg` in the `open` path.
- **`src/factory.rs`** — populated from `template` in the production shard-open path.
- **`src/concurrency.rs`** — consts removed, rationale moved onto two new `ConcurrencyManager` fields; constructor params clamped with `.max(1)` (a `0` batch would stall the drain loop and `.buffered(0)` is invalid).

## Testing

- `direnv exec . cargo build` / `--tests` — clean, no dead-code warnings.
- Existing `grant_scanner_drains_backlog_larger_than_max_per_pass` regression test passes with the preserved 256 default.
- Added config round-trip tests (defaults + custom values) and extended both `*_default_matches_toml_defaults` drift guards.

### DIFF
--- a/benches/bench_helpers.rs
+++ b/benches/bench_helpers.rs
@@ -616,6 +616,8 @@ pub async fn clone_golden_shard(
             hydrate_all_at_startup: false,
             completed_job_expire_s: None,
             terminal_job_expire_s: None,
+            grant_scanner_batch_size: silo::settings::DEFAULT_GRANT_SCANNER_BATCH_SIZE,
+            grant_scanner_buffer_size: silo::settings::DEFAULT_GRANT_SCANNER_BUFFER_SIZE,
         },
         ShardRange::full(),
     )
--- a/src/concurrency.rs
+++ b/src/concurrency.rs
@@ -166,23 +166,6 @@ impl From<crate::codec::CodecError> for ConcurrencyError {
     }
 }
 
-/// Max in-flight `db.get(job_status_key)` lookups during a single
-/// `process_grants` pass. Caps slatedb-side block-fetch fan-out so the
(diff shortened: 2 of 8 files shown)


kirinrastogi:
image

setting these clamps to 4x

```
grant_scanner_buffer_size = 256  # throughput lever (status-lookup concurrency)
grant_scanner_batch_size = 1024  # mostly a memory bound; fewer commits per drain
```

udnay (approved):

PR #376, opened 2026-06-29
Hash siloctl shard-split --at tenant id into the keyspace

scott-rc:
## Why

During the 2026-06-29 silo-staging shard-split gameday, `siloctl shard split --at <tenant>` rejected real tenant ids with errors like `invalid split point: split point 'env-10000468093' is not within shard range [a000..., c000...)`.

The CLI sent `--at` to the server verbatim. The server validates split points against the hex hash keyspace (`ShardRange::contains`, a raw lexicographic comparison) and never hashes the input, so a tenant id never lands inside any hash range. Operators had to work around this by precomputing the boundary with `siloctl tenant hash <tenant>` and passing the hash -- despite the `--at` help text reading "Tenant ID at which to split the keyspace".

## What

- `--at <tenant-id>` now hashes the tenant id client-side (`silo::shard_range::hash_tenant`) before sending, so it works as documented.
- `--at-hash <hash>` added as an escape hatch for passing a precomputed 16-char hex boundary verbatim (validated, normalized to lowercase since the server compares lexicographically). Useful for boundaries that aren't a single tenant.
- `--auto` is unchanged. The three selectors are mutually exclusive.
- Split-point resolution is extracted into a pure helper and covered by unit tests, including a regression test using the gameday's exact range.

This is a UX/correctness fix in the `siloctl` CLI only. It is separate from the shard-split clone data-loss bug in `src/factory.rs` `clone_closed_shard`.

### DIFF
--- a/siloctl/src/lib.rs
+++ b/siloctl/src/lib.rs
@@ -802,12 +802,68 @@ pub async fn heap_profile<W: Write>(
     Ok(())
 }
 
+/// Validate that a string is a raw 16-character hex hash boundary, returning it
+/// normalized to lowercase.
+///
+/// Shard range boundaries are 16-char hex-encoded u64 values compared
+/// lexicographically by the server, so case matters: uppercase hex digits sort
+/// before lowercase ones. `siloctl tenant hash` emits lowercase, so we normalize
+/// to lowercase to keep operator-supplied hashes comparable with the keyspace.
+fn validate_hash_boundary(hash: &str) -> anyhow::Result<String> {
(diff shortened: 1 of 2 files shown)


scott-rc:

- #376 ◀

kirinrastogi (approved):
thanks, I thought I covered this with a get tenant function that computes hash from tenant id, but I missed this here because it doesn't need to look up the tenant, just find the split point. also in my failed shard split I calculated the xxhash ahead of time and used that.

PR #377, opened 2026-06-29
Fix silent data loss when splitting shards on object-store backends

scott-rc:
Splitting a shard on an object-store backend (GCS) silently lost the entire shard's data. The clone step wrote each child's SlateDB manifest into the parent's own per-shard namespace while `open` later read from the child's namespace, so every child opened empty. The loss was deterministic regardless of split point or shard size, and reproduced in staging where a tenant dropped from roughly 8M jobs to a handful. The parent's SSTs were never deleted, so the situation was recoverable -- but every object-store split was a data-loss event.

## Root cause

Object-store backends rooted each shard at its own full prefix and set the db path to the full canonical path, double-nesting data at `<root>/<shard>/<shard>` and -- critically -- leaving the parent store not a shared ancestor of its children. SlateDB reference-clones record parent SST paths relative to one shared store root, so a clone written under the parent's prefix is unreachable from the path `open` resolves for the child. Local-FS avoided this by rooting every shard at the shared prefix before the `%shard%` placeholder; object stores did not.

## Fix

- Resolve object-store backends (Memory/S3/GCS/URL) at the shared placeholder-prefix root with `db_path = shard_name`, symmetric with the local-FS branch. The clone destination for a child now equals the path `open` reads, and the double-nesting is gone. Local-FS resolution is unchanged.
- Share the in-memory backend across shards under one root so multi-shard clone/open is reproducible in tests; each shard previously got a fresh store, so the Memory backend could not host a clone at all.
- Add a defense-in-depth check to the split state machine: before the shard-map commit point, open each cloned child and confirm its whole-shard `total_jobs` matches the parent's pre-close count. A child that opens empty or short aborts the split as a pre-commit failure that recovers the parent, so a latent clone bug surfaces as a failed, recoverable split rather than silent data loss.
- Add object-store coverage for the clone and split paths -- the gap that hid the bug -- including clone round-trips with and without a separate WAL store and a full split state-machine round-trip that asserts every job survives in the child that owns its range.

Object-store shard data is staging-only and disposable, so moving where shards resolve needs no migration.

cc @airhorns

### DIFF
--- a/src/coordination/split.rs
+++ b/src/coordination/split.rs
@@ -517,6 +517,25 @@ impl ShardSplitter {
                             parent_shard_id
                         )))
                     })?;
+
+                    // Capture the parent's whole-shard job count from the still-open
+                    // handle, before closing it. The post-clone verification below
+                    // confirms each child holds this full set (a fresh clone is a
+                    // byte-for-byte copy; cleanup trims each child to its range only
+                    // after commit), catching the empty-child data-loss signature.
+                    let parent_total_jobs = parent_shard
+                        .get_counters()
+                        .await
+                        .map_err(|e| {
+                            SplitExecutionError::PreCommit(CoordinationError::BackendError(
+                                format!(
+                                    "failed to read parent shard {} counters before cloning: {}",
+                                    parent_shard_id, e
+                                ),
+                            ))
+                        })?
+                        .total_jobs;
+
                     parent_shard.close().await.map_err(|e| {
                         SplitExecutionError::PreCommit(CoordinationError::BackendError(format!(
                             "failed to close parent shard before cloning: {}",
@@ -559,7 +578,36 @@ impl ShardSplitter {
                         parent_shard_id = %parent_shard_id,
                         left_child = %split.left_child_id,
                         right_child = %split.right_child_id,
-                        "cloning complete, updating shard map"
+                        "cloning complete, verifying children before commit"
+                    );
+
+                    // Defense-in-depth: before the commit point, open each child
+                    // and confirm it holds the parent's full job set. A child that
+                    // opens empty/short (the object-store resolution bug's data-loss
+                    // signature) fails here, aborting the split as
+                    // PreCommitParentClosed so the shard map stays untouched and the
+                    // parent is recovered/reopenable -- never silent data loss.
+                    self.ctx
+                        .factory
+                        .verify_cloned_children_match_parent(
+                            parent_total_jobs,
+                            &[split.left_child_id, split.right_child_id],
+                        )
+                        .await
+                        .map_err(|e| {
+                            SplitExecutionError::PreCommitParentClosed(
+                                CoordinationError::BackendError(format!(
+                                    "post-clone child verification failed for split of {}: {}",
+                                    parent_shard_id, e
+                                )),
+                            )
+                        })?;
+
+                    info!(
+                        parent_shard_id = %parent_shard_id,
+                        left_child = %split.left_child_id,
+                        right_child = %split.right_child_id,
+                        "children verified, updating shard map"
                     );
 
                     // Pre-add BOTH children to owned set BEFORE updating the shard map.
--- a/src/factory.rs
+++ b/src/factory.rs
@@ -820,9 +820,12 @@ impl ShardFactory {
         )
         .await?;
 
-        let total_jobs = shard.get_counters().await?.total_jobs;
+        // Always close the raw handle, even if the counter read fails, so the
+        // shard's SlateDB instance and its spawned background tasks (grant
+        // scanner, reconcilers) are released -- JobStoreShard has no Drop.
+        let counters = shard.get_counters().await;
         shard.close().await?;
-        Ok(total_jobs)
+        Ok(counters?.total_jobs)
     }
 
     /// Get the database template used by this factory.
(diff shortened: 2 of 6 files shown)


scott-rc:

- #377 ◀

udnay (approved):
This looks good to me, do we have a way to run all tests through a gcs/object store mock? Having different file systems has bitten us more than once I feel and having all tests go through one mock path would help smooth out those inconsistencies wdyt?

scott-rc:
> This looks good to me, do we have a way to run all tests through a gcs/object store mock? Having different file systems has bitten us more than once I feel and having all tests go through one mock path would help smooth out those inconsistencies wdyt?

I'm not gonna lie, I didn't look at the existing code too closely 😅. I assume we already have a mock backend that we use for unit tests, but if we don't that would be a good idea. Although, the reason the bug this PR is fixing exists is because we weren't testing with the real backend, so I would avoid a mock tbh.

udnay:
>  Although, the reason the bug this PR is fixing exists is because we weren't testing with the real backend,

Really? I misread the PR comment, I thought 
> Object-store backends rooted each shard at its own full prefix and set the db path to the full canonical path, double-nesting data at <root>/<shard>/<shard> and -- critically -- leaving the parent store not a shared ancestor of its children

Was because we don't have a mock for GCS and so we run things on local file system and we ended up setting the paths wrong. This is a slatedb issue?

Also to be clear, when I say mock I mean we tell silo/slatedb here's your storage layer `gadgets-cool-gcs-mock` and trust us it's GCS. So that we don't get these weird drifts. This happened before with shard splits where using in-memory file store did not diligently execute the same code paths as we do in prod. We've hit it again, I'm wondering if we ran all of our tests against what slatedb/silo thought was an object store we might expose some more of these flakes.

scott-rc:
> This is a slatedb issue?

No -- slatedb behaved as documented. The bug was in silo's `resolve_at_root` (src/factory.rs). The local-FS branch resolved every shard at the shared storage root with the shard name as the db path (correct). The object-store branch resolved each shard's store at its own full per-shard prefix and then passed the full canonical path as the db path on top of that -- which is what produced the `<root>/<shard>/<shard>` double-nesting and, worse, meant the parent's store was not an ancestor of its children, so clones landed where the child's `open` never looks. Two backends, two different rooting strategies, and only the local-FS one had clone/split coverage.

> Was because we don't have a mock for GCS and so we run things on local file system and we ended up setting the paths wrong.

Close, but the local-FS paths were fine -- the object-store branch was wrong and nothing exercised it. The deeper testability gap: `resolve_object_store(Backend::Memory)` returned a fresh empty `InMemory` on every call, so two shards could never see each other's storage. A multi-shard clone was untestable in-memory by construction.

> when I say mock I mean we tell silo/slatedb here's your storage layer `gadgets-cool-gcs-mock` and trust us it's GCS

That is exactly what this PR sets up. slatedb's `InMemory` implements the same `ObjectStore` interface as the real GCS client, and the Memory backend now (a) shares one store per root and (b) resolves through the same object-store branch as GCS/S3 -- so tests run the identical silo/slatedb code path as prod, minus the network. The new clone and split tests run on it, and the clone round-trip test fails on the pre-fix code. Running the broader suite against it to flush out more of these drifts is a good idea -- it touches most test helpers, so I'd take it as a follow-up PR.

_ Addressed by [Claude Code](https://claude.com/claude-code)_

scott-rc:
> > This is a slatedb issue?
> 
> No -- slatedb behaved as documented. The bug was in silo's `resolve_at_root` (src/factory.rs). The local-FS branch resolved every shard at the shared storage root with the shard name as the db path (correct). The object-store branch resolved each shard's store at its own full per-shard prefix and then passed the full canonical path as the db path on top of that -- which is what produced the `<root>/<shard>/<shard>` double-nesting and, worse, meant the parent's store was not an ancestor of its children, so clones landed where the child's `open` never looks. Two backends, two different rooting strategies, and only the local-FS one had clone/split coverage.
> 
> > Was because we don't have a mock for GCS and so we run things on local file system and we ended up setting the paths wrong.
> 
> Close, but the local-FS paths were fine -- the object-store branch was wrong and nothing exercised it. The deeper testability gap: `resolve_object_store(Backend::Memory)` returned a fresh empty `InMemory` on every call, so two shards could never see each other's storage. A multi-shard clone was untestable in-memory by construction.
> 
> > when I say mock I mean we tell silo/slatedb here's your storage layer `gadgets-cool-gcs-mock` and trust us it's GCS
> 
> That is exactly what this PR sets up. slatedb's `InMemory` implements the same `ObjectStore` interface as the real GCS client, and the Memory backend now (a) shares one store per root and (b) resolves through the same object-store branch as GCS/S3 -- so tests run the identical silo/slatedb code path as prod, minus the network. The new clone and split tests run on it, and the clone round-trip test fails on the pre-fix code. Running the broader suite against it to flush out more of these drifts is a good idea -- it touches most test helpers, so I'd take it as a follow-up PR.
> 
> _ Addressed by [Claude Code](https://claude.com/claude-code)_

Jesus sorry! I had claude code still looking to auto respond to comments on 🤦 

CleanShot 2026-07-02 at 10 02 35@2x

udnay:
All good, so now all of our tests run with silo/slate thinking we are hitting gcs. 👍

kirinrastogi (approved):

PR #384, opened 2026-07-14
Fix shard-split data loss on object-store backends

scott-rc:
Splitting a shard on an object-store backend (GCS in staging/production) silently loses the shard's data: `clone_closed_shard` resolved each shard's store at its own expanded template path, so child clones landed under the parent's per-shard prefix instead of the children's own canonical paths, and the children opened empty after the shard-map commit. A staging split dropped a tenant from roughly 8M jobs to 3 this way. An earlier fix (#377, reverted in #380) made clone destinations equal open paths but moved every existing shard's canonical keys, so deploying it wiped staging a second way.

This PR makes clone destinations equal open paths while keeping every deployed shard's absolute object keys byte-for-byte unchanged, so it deploys with no migration:

- `resolve_at_root`'s object-store branch resolves one shared store at the template root (the prefix before the `%shard%`/`{shard}` placeholder) and computes a layout-preserving database path: the expanded template relative to the root plus the expanded template's store-path component. For the deployed `gs://<bucket>/silo/%shard%` template this reproduces the existing double-nested keys `silo/<shard>/silo/<shard>/...` exactly, while parent and children now share one store -- so the existing clone code becomes correct by construction. Off-boundary placeholders on URL-style backends are rejected loudly instead of silently resolving to different keys.
- The Memory backend shares one in-memory store per root via a process-wide registry and uses the same path arithmetic, so in-memory tests exercise the production key layout. Tests that relied on fresh stores per resolution get unique roots.
- The split state machine verifies both children between clone completion and the shard-map commit: each child's whole-shard job count (an O(1) merged-counter read via a raw, uncached open) must equal the closed parent's, read the same way so the comparison cannot drift against background cleanup. A mismatch aborts the split through the existing parent-recovery path. The markers `cloning complete, verifying children before commit`, `children verified, updating shard map`, and `post-clone child verification failed` are shared constants asserted by tests and intended as runtime probes for deployment verification.
- A debug-build-only fault seam on the factory can skip one child's clone, letting a coordination-level test drive the full abort path (verification fails, shard map untouched, parent recovered and serving). A debug-only armable log-capture layer in `silo::trace` backs the marker assertions.

Regression coverage pins the mechanism itself: the clone round-trip test fails with an empty child if the shared-root arithmetic is reverted to per-shard resolution (verified once during development), and a coordination-level split on the shared-store Memory backend asserts every tenant's jobs survive in the owning child. The standalone compactor keeps reading the same absolute keys, so a lagging compactor image stays compatible.

Related to #377 and #380.

### DIFF
--- a/src/coordination/split.rs
+++ b/src/coordination/split.rs
@@ -517,6 +517,25 @@ impl ShardSplitter {
                             parent_shard_id
                         )))
                     })?;
+
+                    // Capture the parent's whole-shard job count from the still-open
+                    // handle, before closing it. The post-clone verification below
+                    // confirms each child holds this full set (a fresh clone is a
+                    // byte-for-byte copy; cleanup trims each child to its range only
+                    // after commit), catching the empty-child data-loss signature.
+                    let parent_total_jobs = parent_shard
+                        .get_counters()
+                        .await
+                        .map_err(|e| {
+                            SplitExecutionError::PreCommit(CoordinationError::BackendError(
+                                format!(
+                                    "failed to read parent shard {} counters before cloning: {}",
+                                    parent_shard_id, e
+                                ),
+                            ))
+                        })?
+                        .total_jobs;
+
                     parent_shard.close().await.map_err(|e| {
                         SplitExecutionError::PreCommit(CoordinationError::BackendError(format!(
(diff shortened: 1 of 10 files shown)


scott-rc:

- #384 ◀

PR #385, opened 2026-07-15
Delete scheduled jobs in one DeleteJob call

scott-rc:
Claude wanted to add this, but I'm guessing this breaks the alloy spec so you can just ignore this PR if you don't want this.

---

Purging scheduled jobs previously required two RPCs per job: `DeleteJob` refused every non-terminal status with `cannot delete job <id>: job is currently running or has pending requests` -- a message that is false for a `Scheduled` job -- so operators had to send a `CancelJob` first to make each job deletable. A staging fixture cleanup that purged 10,000 inert future-dated jobs paid that cost per job, and the misleading error text made the workaround undiscoverable.

`DeleteJob` now deletes `Scheduled` jobs directly. The scheduled branch runs inside the same serializable-snapshot transaction machinery cancellation uses, re-reads the job status as its first in-transaction operation (the pre-transaction read only routes), and performs the full cleanup a cancel-then-delete pair produces: pending task removal (with the time-0 fallback lookup), concurrency holder and request-record release with post-commit slot regrant and broker-buffer eviction, requester-counter decrement, status-time/metadata index cleanup, attempt-row sweep, and counter bookkeeping without the completed-jobs decrement that only terminal jobs carry. The in-memory queue gauge decrements exactly once per deleted job, with rollback on failed commit attempts so conflict retries cannot drift it. That cleanup logic is shared with the cancel path via one helper so the two cannot silently diverge.

`Running` jobs stay refused, and `JobInProgress` now carries the job's status so the error names what the job is actually doing (`job is Running; cancel it or wait for it to finish`). Terminal-status deletes and idempotent missing-job deletes are unchanged. No proto or siloctl changes -- the CLI's single-ID delete works on scheduled jobs as-is.

Review notes: the scheduled-refusal tests are repurposed as positive coverage (the double-duty pending-request test is split so its Running-refusal coverage survives); counter exactness is asserted mid-purge as well as at the end, since the queue gauge floors at zero and only a partial-purge checkpoint can catch a double decrement; gRPC tests pin the one-call delete (GetJob NotFound, no leasable task) and the status-naming refusal end-to-end.

### DIFF
--- a/src/job_store_shard/cancel.rs
+++ b/src/job_store_shard/cancel.rs
@@ -161,127 +161,12 @@ impl JobStoreShard {
             let job_key = job_info_key(tenant, id);
             if let Some(job_bytes) = try_after_status!(txn.get(&job_key).await) {
                 let job_view = try_after_status!(JobView::new(job_bytes));
-                let task_group = job_view.task_group().to_string();
-                let priority = job_view.priority();
-                let limits = job_view.limits();
-
-                // Locate the pending task by identity (the trailing epoch_ms is
-                // write-only, so we prefix-scan instead of reconstructing the
-                // exact key).
-                let attempt_number = status.current_attempt.unwrap_or(1);
-                let start_time_ms = status.next_attempt_starts_after_ms.unwrap_or(0);
-                let task_found = match try_after_status!(
-                    find_task_by_identity(
-                        &txn,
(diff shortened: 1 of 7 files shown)


scott-rc:

- #385 ◀

PR #386, opened 2026-07-28
Emit silo_leasable_to_start_latency_ms for worker-capacity scaling

scott-rc:
## Why

The silo-worker HPAs scale on `silo_ready_to_start_latency_ms`, which measures from a task's original scheduled start -- deliberately preserved through tenant concurrency-limit chains. A tenant with a deep backlog behind a narrow concurrency queue therefore reports hours-long latencies on every lease, and adding workers cannot reduce them: the fleet pins at maxReplicas while worker CPU sits near-idle. That metric is the right customer-perceived queue-time signal, but the wrong capacity signal.

## What

Adds `silo_leasable_to_start_latency_ms`, a histogram (labels `shard`, `task_group`, same bucket layout as the existing metric) measuring only the time a `RunAttempt` task was leasable -- eligible for a worker to pick up -- before it was leased. It records at exactly one site, the `RunAttempt` lease path, next to the existing metric's recording there.

The leasable moment is derived from the existing task key with no encoding change: `max(start_time_ms, epoch_ms)`. For a future-scheduled task the scheduled start dominates; for a task re-emitted after a concurrency grant the write-time epoch (which is approximately the grant time) dominates. This relies on every `RunAttempt` write site stamping `epoch_ms` at write time or later, which holds across all of them; the one back-dating path (`rate_limit_retry_epoch_ms`) writes only `CheckRateLimit` tasks. The computation is a named function with unit tests pinning both key shapes.

`silo_ready_to_start_latency_ms` is untouched: registration, recorder, and both recording sites are byte-identical. Its docs entries are reworded to state what it actually measures (from the original scheduled start, through concurrency waits) so the two adjacent entries do not contradict.

Downstream, the prometheus-adapter gains a rule for the new metric and the silo-worker HPAs switch to it (separate PRs in global-infrastructure and gadget).

## Review notes

- The exposition test asserts `_count`, `_sum`, and `_bucket` -- the adapter's `seriesQuery` selects on `_bucket`, so its presence is load-bearing.
- The unit tests in `dequeue.rs` cover the epoch-dominant (concurrency-granted), start-time-dominant (future-scheduled), and clamp-to-zero shapes.

### DIFF
--- a/docs/src/content/docs/guides/observability.mdx
+++ b/docs/src/content/docs/guides/observability.mdx
@@ -47,7 +47,8 @@ Leases represent tasks actively being processed by workers.
 | Metric | Type | Labels | Description |
 |--------|------|--------|-------------|
 | `silo_task_leases_active` | Gauge | `shard`, `task_group` | Number of tasks currently leased to workers |
-| `silo_ready_to_start_latency_ms` | Histogram | `shard`, `task_group` | Time between when a task became ready and when it was first leased (in milliseconds) |
+| `silo_ready_to_start_latency_ms` | Histogram | `shard`, `task_group` | Time between a task's original scheduled start and when it was leased, including any wait for concurrency tickets or rate-limit clearance (in milliseconds) |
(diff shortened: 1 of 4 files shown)


scott-rc:

- #386 ◀

PR #387, opened 2026-07-29
Stream grant scanner commits in chunks to cut leasable latency

scott-rc:
cc @airhorns @udnay @kirinrastogi -- I've been working on the silo worker hpas and noticed one of them was unnecessarily maxed out at 80 pods. Fable believes the hpa's [`silo_leasable_to_start_latency_ms`](https://github.com/gadget-inc/silo/pull/386) was inflated because the grant scanner is scanning too much at once, delaying jobs from running when they could have otherwise ran. The solution we went with is to grant more often in between the scan.

---

## Why

Production `silo_leasable_to_start_latency_ms` for the `app-green-all-background-actions-default-spot` task group runs p50 ~1.4-2.9s at ~50 leases/s while the group's 80 workers sit near-idle (~12% CPU) -- the HPA maxed out replicas during a >10x traffic ramp and latency kept climbing, so worker capacity is not the constraint. The latency concentrates on the single shard hosting the hot tenant, whose gating concurrency queue is at capacity: essentially every RunAttempt there flows through the grant scanner (29.5/s `scanned`-path grants vs 2.7/s `immediate`), and the leases in the 1-5s latency band match the scanner grant rate almost exactly.

The mechanism is in `process_grants`: a pass over a backlogged queue pays one object-store round trip per candidate serially (next-hop skip reads inside the scan loop, then downstream floating-limit state reads inside the serial reserve loop) and commits every grant in a single batch at the end of the pass. A granted task therefore becomes leasable only after the whole read-bound pass finishes -- seconds after its slot actually freed -- and its `task_key` epoch is stamped from the invocation's start, which also skews the leasable-latency histogram.

## What

- `process_grants` streams grants out in chunks: every `grant_scanner_commit_chunk_size` grants (new setting, default 16) the accumulated batch is committed durably with a fresh `now_ms`, so early grants in a long pass become leasable while later candidates are still being walked, and holder/epoch timestamps reflect commit time.
- The per-candidate reads are overlapped instead of serial: next-hop skip reads are warmed concurrently for the pass's distinct hops before the skip filter runs, and downstream floating-limit state plus queue hydration are prefetched into a read overlay that the chain resumer's writer serves `get`s from. The overlay holds point-in-time DB reads, matching the writer's existing `get` semantics (batch edits were never visible to it).
- Each chunk commit wakes the granted task groups' brokers through a new `LimitChainResumer::wakeup_task_groups` hook, instead of waiting for the grant scanner's drain iteration to finish every other queue's pass.
- A new `grant_latency` bench (`cargo bench --bench grant_latency`) reproduces the pathology and gates the fix: it injects 20ms object-store read latency, saturates a gate queue whose waiters each carry a distinct downstream floating hop (keys padded so each state row occupies its own SST block), keeps the request front memtable-warm like a continuously converting queue, and asserts per-grant time-to-leasable, end-to-end release-to-lease, and invocation-duration thresholds.

## Benchmark results

64-grant saturated queue, 20ms injected GET latency:

| Metric | main | this branch | threshold |
|---|---|---|---|
| grant -> leasable p50 | 2.81s | 87ms | 300ms |
| grant -> leasable p99 | 2.81s | 89ms | 800ms |
| `process_grants` invocation | 2.79s | 81ms | 2.5s |
| release -> lease, unsaturated path | 54ms | 55ms | 250ms |

The baseline is close to the production signature (hot shard p50 2.9s; healthy shards ~35ms), so the expectation is the production p50 drops into the low hundreds of ms. The remaining `ready_to_start` gap is genuine capacity wait at the app's concurrency limit and is unaffected by design.

## Testing

- `cargo test` main suite: 925 passed; the 9 failures are `rate_limit_tests` that require a locally running Gubernator and fail at connection setup before reaching this code.
- DST turmoil suite (`--features dst`): 32/32 passed, including the grant scanner scheduling scenario.
- The new bench fails on `main` and passes with this branch; the green result reproduced across runs within 3ms.

Two intentional behavior-parity notes: next-hop-skipped candidates now occupy pass slots (they consume the scan budget exactly as before), and a cross-chunk read of floating state through the prefetch overlay matches the within-pass staleness the batch write already had.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -166,6 +166,10 @@ required-features = ["server"]
 name = "task_scanner_profile"
 harness = false
 
+[[bench]]
+name = "grant_latency"
+harness = false
+
 [[test]]
 name = "turmoil_runner"
 path = "tests/turmoil_runner/main.rs"
--- a/benches/grant_latency.rs
+++ b/benches/grant_latency.rs
@@ -1,0 +1,698 @@
+//! Per-grant time-to-leasable latency through the concurrency grant scanner
+//! under object-store read latency.
+//!
+//! Models the production shape where a single tenant's gating queue is
+//! saturated: every RunAttempt flows through a deferred concurrency request
+//! that `process_grants` drains, and each chain-resume read pays an
+//! object-store round trip (cold block cache over GCS).
+//!
+//! Cases:
+//! 1. `grant_to_leasable` — free the gate's capacity, run one
+//!    `process_grants(GRANT_COUNT)` invocation while a concurrent dequeuer
+//!    timestamps each granted task as it becomes claimable. The p50/p99 of
+//!    (lease time - invocation start) is the metric the production
+//!    `silo_leasable_to_start_latency_ms` histogram degrades on.
+//! 2. `release_to_lease` — with the grant scanner running, release one holder
+//!    and measure until the next waiter is actually leased. Covers the
+//!    release -> nudge -> grant -> broker-wakeup -> claim pipeline.
+//! 3. Invocation-duration guard — total `process_grants` wall time from
+//!    case 1 must not regress; latency fixes must not trade away throughput.
+//!
+//! Thresholds are asserted at the end of the run; a failing run prints all
+//! measurements first.
+
+use async_trait::async_trait;
+use futures::stream::BoxStream;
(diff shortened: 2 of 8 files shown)


scott-rc:

- #387 ◀

PR #388, opened 2026-07-31
Fix msgpack decode failures for reaped attempt error bytes

scott-rc:
The Gadget editor's background actions page intermittently fails with "[GraphQL] Data read, but end of buffer not reached 108" (~30-170 occurrences/day across 15+ environments). Root cause is a writer/reader defect pair around attempt error bytes:

- The lease-expiry reaper stores its "lease expired at ... worker=..." message as raw UTF-8 in a field every reader decodes as msgpack. The first byte of "lease..." (0x6C = 108) decodes as the integer 108 with trailing bytes, so msgpackr throws.
- The TS client fails the entire getJob call when any single attempt blob is undecodable, so one corrupt attempt breaks the whole background-actions listing.

This PR fixes both ends:

- **Server**: `reap_expired_leases` now encodes the error message as a msgpack string via `rmp::encode::write_str`. An audit of the other `Encoding::Msgpack` construction sites found no other writer producing non-msgpack bytes (worker-reported outcomes pass client-encoded msgpack through untouched; the query path hand-encodes via `record_batches_to_msgpack`).
- **TS client**: attempt `result`/`errorData`, the job-level `getJob` result, and the `getJobResult` data decodes go through a new exported `tolerantDecodeAttemptBytes` helper: valid msgpack decodes as before, undecodable bytes fall back to their UTF-8 string, and empty/absent bytes yield `undefined` (zero-length blobs exist at rest and previously threw "No bytes to decode"). Job payload decodes and `decodeBytes` itself stay strict, and `JobAttempt.errorData` widens to `unknown` to admit the string fallback.

Corrupt blobs already at rest are handled by the reader fallback and age out with job retention; no data migration. Gadget's `parseFailureReason` already renders a plain-string `errorData` as `{ code, message }`, so reaped attempts show a readable failure reason in the editor.

Tested with a Rust regression test decoding a reaped attempt's error bytes as a msgpack string, unit tests for the tolerant helper, and mocked-transport `getJob`/`getJobResult` tests covering a corrupt attempt among valid ones, zero-length result bytes, and raw-UTF-8 failure data.

### DIFF
--- a/src/job_store_shard/lease.rs
+++ b/src/job_store_shard/lease.rs
@@ -584,15 +584,20 @@ impl JobStoreShard {
                 // [SILO-REAP-3][SILO-REAP-4] Report as worker crashed
                 // SILO-REAP-3: Post: Set job status to Failed (via report_attempt_outcome)
                 // SILO-REAP-4: Post: Set attempt status to AttemptFailed
+                let message = format!(
+                    "lease expired at {} (now {}), worker={}",
+                    decoded.expiry_ms(),
+                    now_ms,
+                    decoded.worker_id()
+                );
+                // Readers treat attempt error bytes as msgpack, so encode the
+                // message as a msgpack string rather than raw UTF-8.
+                let mut error = Vec::new();
+                rmp::encode::write_str(&mut error, &message)
+                    .expect("msgpack encoding into a Vec cannot fail");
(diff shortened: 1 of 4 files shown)


scott-rc:

- #388 ◀

PR #389, opened 2026-08-05
Plan multi-status jobs queries as unions of status index range scans

scott-rc:
cc @airhorns @udnay @kirinrastogi -- claude believes this will help prevent timeouts on the queues and admin pages.

---

Multi-status predicates on `status_kind` -- the shapes the gadget admin and IDE background-action pages emit (`status_kind IN (...)`, OR-of-equalities, and OR-of-AND state filters) -- fell through the jobs-table planner to `FullScan`. On a large tenant (~380k jobs) that streams the entire job-info range regardless of how few rows match, blowing the 5s statement timeout. This makes those shapes scale with matching rows instead of tenant size, with no client-visible protocol or schema changes.

## Planner

- New tenant-gated `StatusUnion` strategy: literal `IN`-lists (parsed directly, since only lists of 3 or fewer lower to ORs), OR-of-equalities, and OR-of-AND shapes where every arm pins `status_kind` (pins may be equalities, `IN`-lists, or OR-of-equalities). Tenant-less multi-status predicates, `NOT IN`, arms without a status pin, and lone equalities keep their existing strategies.
- Filter classification now sees whole expressions instead of a single column name. Only the predicate driving the scan is `Exact`: pure `IN`/OR-of-equality unions that match the scanned statuses suppress the redundant post-filter, while OR-of-AND residue shapes, ANDed extra status predicates, and status predicates under a metadata strategy all stay `Inexact` so DataFusion re-applies them. Tenant/id equalities are `Exact` only when their value matches the planned strategy, so conflicting duplicates are post-filtered rather than silently dropped.

## Execution

- Unions run as sequential per-status index range scans in both the point-lookup hydration path and the index-only path, with all arms' ranges planned from a single clock snapshot so the virtual Waiting/Scheduled split partitions the stored `Scheduled` range without overlap, and emitted labels stay consistent with the arm that produced them. Scan limits short-circuit across arms.
- The index-only projection path (aggregates, `GROUP BY status_kind` counts) supports unions directly, so those shapes do not regress to per-row status lookups. The counter-served `COUNT(*)` fast path deliberately stays FullScan-only.
- A new `silo_query_point_lookups_total` per-shard counter makes the index-only vs hydration choice observable; tests use it to pin the no-regression property.

## Tests

Strategy/classification tests cover each recognized and rejected shape; integration tests on seeded shards assert hand-specified row sets (including the virtual Waiting/Scheduled mapping and OR-of-AND residue filtering), scanned-keys deltas bounded by matching entries rather than tenant size, and limit short-circuiting across union arms.

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -142,6 +142,10 @@ pub struct Metrics {
     /// stops early (LIMIT reached, deadline fired, stream dropped) stops
     /// advancing this counter — the signal that a query is not running as a zombie.
     query_scanned_keys: CounterVec,
+    /// Rows hydrated via job_info/status point-lookups by query scans, per
+    /// shard. Stays flat for index-only scans — the signal that an aggregate
+    /// shape rode the index-only path instead of per-row hydration.
+    query_point_lookups: CounterVec,
 
     // Shard/broker metrics
     shards_owned: Gauge,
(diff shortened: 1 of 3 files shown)


scott-rc:

- #389 ◀

PR #390, opened 2026-08-10
Fix split-child cleanup stranding on non-initiator nodes

scott-rc:
## Problem

After a shard split, a child's defunct-data cleanup only runs if `CleanupPending` exists in its database -- and that status was written post-commit, only for children the split initiator ended up owning itself. A child acquired by a different node had no status key at all, and the absent key reads as `CompactionDone` ("nothing to do"), so the acquiring node silently skipped cleanup and the child retained every key outside its range indefinitely. The 2026-08-10 cleanup gameday on silo-staging hit exactly this: child `2c04b378` was acquired by a non-initiator node, never ran cleanup, and still holds ~581k defunct jobs. The initiator's own children were exposed too: a fast shard-guard acquisition could read the absent key in the window between the post-commit open and the status write.

## Fix (three cooperating changes)

1. **Initialize cleanup metadata at clone time.** Both children get `CleanupPending` -- with inherited progress, legacy complete marker, and completion timestamp cleared, since a clone is byte-for-byte and inherited terminal state would short-circuit cleanup as "already complete" -- inside the clone-verification pass, before the shard-map commit, while the initiator exclusively holds both cloned databases. An explicit MemTable-type flush lands the write in shared storage rather than only the initiator's node-local WAL. Initialization failure aborts the split pre-commit under its own log marker. The post-commit per-owner status write is removed, which also closes the acquire-before-status race.
2. **Status plumbing distinguishes "absent" from `CompactionDone`.** A raw reader returns `Option<SplitCleanupStatus>`; the defaulted reader keeps its name and semantics for display callers. A first write into a shard with no status key is a valid initial transition; the regression WARN now fires only for a genuine backward transition between two present statuses, so a normal split's happy path is WARN-free.
3. **Auto-backfill stranded children on acquisition.** The background-cleanup spawn takes the shard map's parent record alongside the range. When the database has no status key: a recorded parent with no completion timestamp means the child was stranded by the pre-fix path, so the acquirer resets its cleanup metadata (logged at info) and cleans it; a present completion timestamp skips with an info log; no recorded parent is an original shard and stays untouched. Any present status follows `needs_work()` exactly as before. All five coordinator spawn sites pass the metadata, so existing stranded children (staging's included) heal on their next acquisition without operator action.

Known limit: a child stranded pre-fix whose parent had completed its own cleanup inherited a present terminal status from the clone, so the absent-status signal cannot identify it; no such shard is known to exist today.

## Testing

- Guard-free mock-split harness asserts both children carry exactly `CleanupPending` with fully reset metadata pre-commit, with no coordinator or shard guards racing the value, read through a non-registering factory helper.
- Cross-node visibility: a passive second factory sharing the data root but carrying its own WAL root reads the initialized metadata -- fully shared or fully per-node roots would make this assertion vacuous.
- End-to-end two-node etcd test: bounded-load rendezvous caps each node at one child, so the non-initiator deterministically acquires a child the initiator never opened and its cleanup deletes the defunct half of the parent's data.
- Second-generation split: grandchildren of a cleanup-complete child get fresh metadata and their cleanup actually deletes keys; the whole flow emits no regression WARN.
- Init-failure abort (via a debug-only fault seam), the full backfill decision table including the inherited-complete-progress shape, and registry non-poisoning are each covered; the spawn returns its task handle so tests await decisions instead of sleeping.

### DIFF
--- a/src/coordination/etcd.rs
+++ b/src/coordination/etcd.rs
@@ -401,7 +401,7 @@ impl EtcdCoordinator {
                 }
             };
 
-            shard.maybe_spawn_background_cleanup(range, parent_shard_id);
+            let _ = shard.maybe_spawn_background_cleanup(range, parent_shard_id);
 
             {
                 let mut owned = self.base.owned.lock().await;
@@ -1151,10 +1151,8 @@ impl EtcdShardGuard {
 
                                     // Spawn background cleanup if this shard has pending cleanup work
                                     // (e.g., it's a split child or was re-acquired after a crash)
-                                    shard.maybe_spawn_background_cleanup(
-                                        range.clone(),
-                                        parent_shard_id,
-                                    );
+                                    let _ = shard
+                                        .maybe_spawn_background_cleanup(range, parent_shard_id);
 
                                     {
                                         let mut st = self.ctx.state.lock().await;
--- a/src/coordination/etcd.rs
+++ b/src/coordination/etcd.rs
@@ -401,7 +401,7 @@ impl EtcdCoordinator {
                 }
             };
 
-            let _ = shard.maybe_spawn_background_cleanup(range, parent_shard_id);
+            drop(shard.maybe_spawn_background_cleanup(range, parent_shard_id));
 
             {
                 let mut owned = self.base.owned.lock().await;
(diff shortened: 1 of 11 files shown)


scott-rc:

- #390 ◀

PR #392, opened 2026-08-19
Sample the coordinator's owned set into silo_shards_owned

scott-rc:
## Why

During the placement-rings gameday on silo-staging (2026-08-19), `silo_shards_owned` read 0 on every pod, including pods owning shards. The gauge is registered, but `Metrics::set_shards_owned` has no production caller, so any dashboard or alert built on it reflects nothing. This is phase 1 of hardening placement rings: make what a node owns visible, without touching the assignment algorithm or the coordinator backends.

## What

- New `placement_metrics` module: a single-pass `sample` that sets `silo_shards_owned` from an owned shard list, and a `run_sampler` driver that reads `owned_shards()` from the `Coordinator` trait object on a fixed interval and exits on the shutdown broadcast.
- `main` spawns the driver beside the tokio-runtime metrics scraper, only when metrics are enabled, and awaits its join handle at shutdown.
- The gauge's HELP text states what it counts (shards this node has acquired and opened -- the owned set is inserted into only after a successful open on every backend) and that it is sampled every 5s, so the staleness bound is visible to anyone building on it.

## Review notes

- `src/coordination/` is untouched; the sampler only reads through the trait.
- Tests (`tests/placement_metrics_tests.rs`) assert the gauge by gathering the registry after a pass over a hand-built owned list, so they need no coordinator.
- Later PRs in this stack add the per-ring unassigned gauge and strand logging on top of this sampler.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,6 +20,7 @@ pub mod job_store_shard;
 pub mod keys;
 pub mod metrics;
 pub mod pb_convert;
+pub mod placement_metrics;
 #[cfg(feature = "server")]
 pub mod query;
 pub mod retry;
--- a/src/main.rs
+++ b/src/main.rs
@@ -208,6 +208,17 @@ async fn main() -> anyhow::Result<()> {
         ))
     });
 
+    // Periodic placement sample: reports the coordinator's owned shard set so
+    // dashboards and alerts see what this node actually owns.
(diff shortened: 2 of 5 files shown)


scott-rc:

- #392 ◀
    - #393
        - #394
            - #395
                - #396
                    - #397

PR #395, opened 2026-08-19
Let siloctl configure an unowned shard and explain the dead end

scott-rc:
## Why

In the silo-staging gameday the command that would have recovered the stranded shard -- `siloctl shard configure <shard>` with no ring -- failed with `Error: invalid format`. `connect_to_shard_owner` resolves the shard's owner to an empty address, builds `http://`, and dies before any RPC leaves the process, although the server-side handler has no ownership requirement. The message says nothing about ownership, so an on-call engineer reads it as a malformed argument. Recovery was a hand-edited ConfigMap.

## What

- `connect_to_shard_owner` distinguishes "shard not in cluster" (existing error, unchanged) from "shard has no owner" (empty `grpc_addr` in the cluster-info response) and returns a typed `ShardUnownedError` built from that same response: shard `<id>` has no owner; its placement ring `<ring>` has no members (members advertise: `<rings, or default>`).
- `shard_configure` alone catches that error and opens a fresh connection to `--address`, since the server's `configure_shard` handler does not require ownership -- so `siloctl shard configure <stranded-shard>` is the one-command recovery. Every other shard-routed command (query, job, split, ...) surfaces the message as its error.

## Review notes

- Tests use an in-test `Coordinator` stub whose owner map comes from the shared `compute_shard_owner_map` over a member list advertising only the default ring and a shard map with one shard pinned to a named ring, so the cluster-info response carries a genuinely ownerless shard.
- The fallback route still reaches the server's ring validation: configuring a stranded shard onto another unpopulated ring without the override is refused with the server's message.

### DIFF
--- a/siloctl/src/lib.rs
+++ b/siloctl/src/lib.rs
@@ -56,6 +56,39 @@ fn ensure_http_scheme(addr: &str) -> String {
 
 type AuthSiloClient = SiloClient<InterceptedService<Channel, AuthInterceptor>>;
 
+/// How the default placement ring is named in the no-owner error.
+const DEFAULT_RING: &str = "default";
+
+/// The shard exists but the cluster's desired assignment gives it no owner:
+/// no current member is in its placement ring.
+///
+/// Commands that need the owner surface this as their error; `shard configure`
+/// catches it and sends the request to the connected node instead, which is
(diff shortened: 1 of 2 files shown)


scott-rc:

- #393
    - #394
        - #395 ◀
            - #396
                - #397

PR #397, opened 2026-08-19
Harden the placement-ring changes after review

scott-rc:
## Why

A review of the placement-ring stack (#392-#396) found no blocking issues but a handful of robustness gaps that are cheaper to land together than to fold back into each PR.

## What

- Placement sampler: `silo_shards_owned` is set even when the owner-map read fails (the owned set does not depend on it), and the owner-map read is raced against shutdown so a hung coordination backend cannot delay process exit. The sampler clears a ring's series once per ring rather than once per shard.
- `RoutingClient::random_shard` draws in a single pass with a direct override lookup instead of rescanning the owner list per shard under the topology lock.
- `siloctl`'s no-owner error renders `(none)` when the cluster-info response lists no members.
- Housekeeping: the sampling interval constant lives beside the gauge registrations in `metrics`; the `UnassignedShards` alias lives in `coordination` beside its producer; `DEFAULT_RING_LABEL` documents that a shard explicitly pinned to a ring literally named `default` renders under the same label.

## Tests added

Owner-map failure path for the sampler; one ring clearing while another stays unassigned; `all_addresses` dropping empty owners and random draws skipping unowned shards; a downcast assertion that the routing client's error is the typed `ShardUnownedError`; a stronger assertion that `siloctl`'s configure fallback reaches the server's validation rather than the client-side error.

### DIFF
--- a/siloctl/src/lib.rs
+++ b/siloctl/src/lib.rs
@@ -77,12 +77,15 @@ pub struct ShardUnownedError {
 
 impl std::fmt::Display for ShardUnownedError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let advertised = if self.advertised_rings.is_empty() {
+            "(none)".to_string()
+        } else {
+            self.advertised_rings.join(", ")
+        };
         write!(
             f,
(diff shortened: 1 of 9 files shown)


scott-rc:

- #393
    - #394
        - #395
            - #396
                - #397 ◀

PR #398, opened 2026-08-20
Make concurrency test dequeues robust to broker buffer scan lag

scott-rc:
The grant-scanner concurrency tests collect their initial holder tasks assuming every `dequeue` call makes visible progress: `grant_scanner_live_headroom_reserves_slots_for_immediate_grants` asserted that a single call returns all 10 granted tasks, `grant_scanner_handles_backlog_larger_than_status_lookup_concurrency` and `grant_scanner_drains_backlog_larger_than_max_per_pass` asserted every iteration of their drain loop returns a non-empty batch, and `grant_scanner_full_queue_cleans_stale_bounded` and `grant_scanner_caps_grants_at_free_capacity` looped with no bound at all. The broker buffer is filled asynchronously by a background DB scan, and `claim_ready_or_nudge` in `src/task_broker.rs` polls for only ~25ms before returning an empty batch -- so under heavy parallel `cargo test` load a dequeue can legitimately come back partial or empty even though the tasks exist in the DB. The headroom test failed intermittently this way when the full suite ran in parallel, and the unbounded loops would spin forever if a task were genuinely lost.

This adds a `dequeue_task_ids_until` helper to the shared test helpers that accumulates task ids across repeated `dequeue` calls until the expected count arrives, bounded by a 30-second deadline that fails loudly with the count collected so far, and routes all five call sites through it.

Verified with the full `job_store_shard_concurrency_tests` suite plus repeated runs of the affected tests under six-way concurrent CPU contention.

### DIFF
--- a/tests/job_store_shard_concurrency_tests.rs
+++ b/tests/job_store_shard_concurrency_tests.rs
@@ -2110,14 +2110,7 @@ async fn grant_scanner_handles_backlog_larger_than_status_lookup_concurrency() {
             .await
             .unwrap();
     }
-    let mut holder_tasks = Vec::with_capacity(N);
-    while holder_tasks.len() < N {
-        let tasks = shard.dequeue("w", "tg", N).await.unwrap().tasks;
-        assert!(!tasks.is_empty(), "dequeue should return holders");
-        for t in tasks {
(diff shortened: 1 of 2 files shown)


scott-rc:

- #398 ◀

PR #400, opened 2026-08-20
Detect lease overwrites at RunAttempt dispatch time

scott-rc:
## Why

A production incident showed the silo server dispatching the same RunAttempt task twice ~927ms apart. The second dispatch silently overwrote the lease record at the same `leased_task_key`, so nothing observable fired at the moment the anomaly happened -- localizing it took a multi-hour trace investigation. The suspected source is the ticket-grant path that materializes RunAttempt tasks carrying the enqueue-time task id; finding that root cause needs data.

## What

Adds a detection guard to `handle_run_attempt`: before writing the lease, it point-reads the task's lease key, and when a still-live lease is found it emits a warn event (task id, job id, tenant, both worker ids, remaining TTL) and increments `silo_task_lease_overwrites_total{shard,task_group,source="stored"}`. Same-iteration lease writes sit in the uncommitted batch where the read cannot see them, so the iteration also tracks its leased task ids and counts a repeat under `source="batch"`. The two labels localize which path produced the overwrite.

Detection only: dispatch proceeds unchanged, the guard adds no writes and at most one point read to a path that already point-reads job info, and no guard failure (read or decode error) can alter a dispatch outcome. Prevention is deliberately deferred until this counter establishes which paths actually hit it -- refusing dispatch on a false positive would drop work, which is worse than the duplicate execution the client now contains on its side.

Also pins the exact gRPC status messages "lease not found" and "lease owner mismatch" with heartbeat integration tests, since deployed TypeScript clients match those strings to classify a lost lease; a wording change would silently downgrade lease-lost to a transient error.

## Review notes

The three tests in `job_store_shard_dequeue_tests.rs` (stored overwrite, batch overwrite, expired lease stays silent) are the behavioral spec. The metric is documented in the observability guide. The client-side counterpart (worker hardening) is #399.

### DIFF
--- a/docs/src/content/docs/guides/observability.mdx
+++ b/docs/src/content/docs/guides/observability.mdx
@@ -51,6 +51,7 @@ Leases represent tasks actively being processed by workers.
 | `silo_leasable_to_start_latency_ms` | Histogram | `shard`, `task_group` | Time a task spent leasable — eligible for a worker to pick up — before it was leased, excluding waits for concurrency tickets or rate-limit clearance (in milliseconds) |
 | `silo_lease_reaper_duration_seconds` | Histogram | `shard` | Duration of expired lease reaper scan operations |
 | `silo_lease_reaper_scans_total` | Counter | `shard` | Total number of expired lease reaper scan operations |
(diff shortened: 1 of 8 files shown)


scott-rc:

- #400 ◀

PR #402, opened 2026-09-02
Serve enqueue_time_ms from the status index for large-tenant listings

scott-rc:
## Why

The IDE queues page lists a tenant's jobs newest-first by `enqueue_time_ms`. That column lived only in the `JOB_INFO` record, so any query projecting it forced the jobs scanner onto a per-row `JOB_INFO` decode (the `job_info`/`job_status` merge join for a tenant scan, batched point-lookups for status-filtered scans). On tenants with hundreds of thousands of jobs the listing query exceeds the statement timeout, while the same shape projecting only index-served columns completes comfortably. This PR makes `enqueue_time_ms` an index-served column.

## What changes

The five commits are meant to be read in order; each stands on its own.

1. **Status record and index entry carry the enqueue time.** The `JobStatus` flatbuffer gains a nullable `enqueue_time_ms` (appended, so existing records decode with `None`), and the status/time index entry's value, empty until now, carries the same value as a big-endian i64. One counter-free row writer produces both rows so they cannot drift; the counter-merging wrapper calls it. Transitions copy the value forward from the incoming status, the old status record, or a `JobView` already in scope, and fall back to at most one `JOB_INFO` read, counted on `silo_enqueue_time_repair_reads_total`. The shard gains a backfill-completion flag mirrored by a marker key.
2. **One-shot backfill sweep.** A per-shard background task, behind the new `[database.enqueue_time_backfill]` setting (off by default), walks `JOB_STATUS` in batches, fills the value into rows that lack it inside a conflict-detecting transaction, keeps each row's `expire_ts` exactly, moves no counters, checkpoints progress so a restart resumes, and sets the completion marker when done. A row a concurrent transition rewrites is skipped, because that transition wrote the value itself.
3. **Query engine.** Projection analysis makes one decision: `enqueue_time_ms` is index-served exactly when the status-index fast path is entered with it projected. Tenant-scoped `Status` and `StatusUnion` strategies take the index path regardless of the marker (their alternative already reads `JOB_INFO` per row); a tenant-scoped `FullScan` takes it once the shard's marker is set. Entries whose value is missing are hydrated individually from `JOB_INFO` and counted on `silo_enqueue_time_fallback_hydrations_total`. `EXPLAIN` now names the path taken (`status-counters`, `status-index`, `fullscan-join`, `pairs`), and the describe hook applies the same dispatch as the scan.
4. **Benchmark.** The query benchmark suite gains the editor's listing shape, measured on a clone of the golden shard in three states (marker set, marker cleared, index values stripped). The golden shard metadata carries a format version so a cached dataset from before the schema change is regenerated rather than reused.
5. **Review hardening.** The sweep skips undecodable rows and restarts from unreadable progress instead of wedging, checks cancellation per row, and writes its final progress and marker atomically; tests widen to cover hydration on every index-served shape and a 120k-job tenant.

## Rollout

Old and new nodes stay correct in both directions: an old node writes empty index values and status records without the field, which a new node treats as "unknown" and hydrates; an old node ignores the value and the appended field. Enabling `enqueue_time_backfill` in the deployment configuration is the rollout step for existing shards; nothing in this PR turns it on.

## Reviewing

Docs updated: the querying guide lists which `jobs` columns are index-served, and the server-configuration reference documents the sweep settings. The 120k-job test is `#[ignore]` because its import takes minutes; run it with `cargo test --test query_index_enqueue_time_tests -- --ignored`.

### DIFF
--- a/benches/bench_helpers.rs
+++ b/benches/bench_helpers.rs
@@ -48,6 +48,11 @@ pub const DEEP_QUEUE_WAITERS: usize = 20_000;
 
 const METADATA_FILE: &str = "golden-metadata.json";
 
+/// On-disk format of the golden shard. Bump whenever the rows a generated
+/// shard contains change shape (a schema field, an index value encoding);
+/// a cached shard with a different version is removed and regenerated.
+pub const GOLDEN_FORMAT_VERSION: u32 = 2;
+
 // ---------------------------------------------------------------------------
 // Metadata
 // ---------------------------------------------------------------------------
@@ -55,6 +60,8 @@ const METADATA_FILE: &str = "golden-metadata.json";
 /// Persisted metadata about the golden shard, written as JSON.
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 pub struct GoldenShardMetadata {
+    /// Must equal [`GOLDEN_FORMAT_VERSION`] for the cached shard to be reused.
+    pub format_version: u32,
     pub total_jobs: usize,
     pub checkpoint_id: Uuid,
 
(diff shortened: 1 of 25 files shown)


scott-rc:

- #402 ◀

PR #406, opened 2026-09-15
Sweep orphaned concurrency holders and record holder owners

scott-rc:
## Why

A durable concurrency holder row can outlive the task it was granted for: when a lease is reaped and nothing ever reports an outcome for the task, the holder row stays behind and occupies one of the queue's `max_concurrency` slots forever. Every later job that requests the queue parks behind it, and once enough holders leak the queue is wedged until an operator drops the tenant's holders by hand. Silo already self-heals in-memory ghosts and holders that a worker reports late, but nothing covered a durable holder whose task is simply gone. The holder record also stored only a grant timestamp, so the `queues` SQL table could not say which job owned a slot.

## What

- Holder records carry the owning `job_id` and total `attempt_number` at every grant path (grant scanner, dequeue-time ticket conversion, immediate reserve). Records without owner fields decode with them absent, so rows written by an earlier build stay readable, and the `queues` table populates `job_id` for holder rows that carry an owner.
- A pure classifier maps a holder record, whether an unexpired lease exists for its task, and the owning job's status to a live-or-orphan verdict with a reason (`job_missing`, `job_terminal`, `running_without_lease`, `attempt_superseded`, `stale`). An expired lease row still awaiting the reaper counts as absent.
- On the existing periodic concurrency-reconcile tick, each shard walks its durable holder prefix in bounded slices with an in-memory cursor, classifies holders past a grace window, and purges a holder only after two consecutive full passes classify it as an orphan, so a snapshot racing an in-flight grant or lease write cannot release a live slot. Purging hydrates the queue, point-reads each holder row, deletes the rows that exist in one durable batch, then releases the in-memory reservations and wakes the grant scanner. Each purge logs at `warn` with tenant, queue, task id, owner, and reason, and increments `silo_concurrency_orphan_holders_purged_total{shard, reason}`.
- Three database settings tune the sweep: `orphan_holder_sweep_slice` (default 5000), `orphan_holder_grace_ms` (default 60000), and `orphan_holder_stale_ms` (default 86400000). The stale rule purges any lease-less holder older than the threshold regardless of owner, which is what recovers holders written without owner information; setting it to 0 disables that rule. The internals guide and the server-configuration reference describe the sweep and the settings.

## Reviewing

Start with `classify_orphan_holder` in `src/concurrency.rs` and its decision-table test, then `purge_orphan_holders` in `src/job_store_shard/lease.rs` for the purge ordering, then `src/job_store_shard/orphan_sweep.rs` for the sliced walk and two-pass confirmation. The integration tests in `tests/holder_leak_tests.rs` include a reproduction of the production wedge: a cap-5 queue with six stale owner-less holders and a parked requester that is admitted once the sweep runs. The stale rule accepts the same over-admit-by-one tradeoff the web UI's Drop holders action makes, applied to one holder at a time.


scott-rc:

- #406 ◀

